PR #9, opened 2021-12-13
Add known value tests for fixed-size P-384 functions

jargh:
The new file "known_value_tests_p384.h" contains 1900 known
value tests (generated independently via the OCaml Num library)
for all the fixed-size P-384 functions (i.e. all other than the
arbitrary-size modulus functions). Two groups are selected for
the endian-ness of the machine, with different known answers in
the two cases (in general).

The new Makefile target "complete" runs all these known value tests
for P-384 functions *and* the usual 100 dynamically generated tests
for all functions, without filtering output and giving an explicit
return code to indicate success (0) or failure (nonzero = number of
test case failures).

*Issue #, if available:*

*Description of changes:*


dkostic (commented):
Looks good to me! It might be good to add the script that generates the tests, maybe in a comment block at the top `known_value_tests_p384.h`?

SalusaSecondus (approved):
LGTM, but I agree that including the generating script (either as a comment or a stand-alone file in the `/tests` directory) would be a good thing.

I'll fix up the `buildspec` file when this is merged.

jargh:
Makes sense; now added the test-generating OCaml program as well.

SalusaSecondus (approved):
LGTM

dkostic (approved):

PR #11, opened 2021-12-13
Actually check results of proofs

SalusaSecondus:
Actually check results of proofs.

**We must wait for all four checks to complete before merging this PR. That is expected to take about 3 hours.**


SalusaSecondus:
Proof checks are passing. One more change needs to be made to the build spec but it doesn't impact these tests so as soon as all checks pass here I'll make it and then we can review and merge.

jargh (approved):

PR #13, opened 2021-12-14
secp256k1 modular multiplication and squaring

jargh:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/arm/Makefile
+++ b/arm/Makefile
@@ -211,6 +211,7 @@ OBJ = fastmul/bignum_emontredc_8n.o \
       p521/bignum_tomont_p521.o \
       p521/bignum_triple_p521.o \
       secp256k1/bignum_add_p256k1.o \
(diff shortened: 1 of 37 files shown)


nebeid (approved):
I didn't go into the files, but here is the offline conversation with John:
Nevine Ebeid 
Are the new operations part of CI? I think yes for the proofs.
Do they have know-answer tests?
Is the CI gating the merge or you prefer not to because it takes multiple hours?

John Harrison 
Yes, the CI should be applied for both (dynamically generated) tests and proofs. It's basically doing "make complete" in the tests dir and "make proofs" in the arm and x86 dirs. But indeed, since the proofs take a long time it was decided not to gate merges on proofs, only tests. For known answer tests, that is only for the P-384 functions at the moment though the infrastructure could easily be used to add more.

PR #16, opened 2021-12-17
Extend and improve constant time instruction checking

jargh:
The ctCheck test now includes the secp256k1 functions (with a few new
instructions in the allowed list for ARM) and is intended to be more
robust against different variants of objdump (e.g. those found on Mac
OS which format things slightly differently). The allowed x86 instructions
have been augmented with variants using "l" or "q" size suffixes where
necessary, since objdump versions differ in whether they produce them.

*Issue #, if available:*

*Description of changes:*


tests/Makefile
```diff
@@ -18,8 +18,6 @@
 # If this block fails to do it automatically, replace with manual
 # setting: ARCH=x86 or ARCH=arm as appropriate
 
-.PHONY: clean ctCheck
```
SalusaSecondus, line 21: This should probably remain. It tells `make` that these targets need to be run whenever they are requested and must never be considered "fresh". Otherwise it is possible for `make` to think that the tests have already run, not become stale, and thus not re-check the output.

jargh: OK, thanks: now the .PHONY target should be restored.

SalusaSecondus (approved):
LGTM

PR #19, opened 2021-12-25
Add alternative variants of secp256k1 functions

jargh:
This is adding alternative-microarchitecture versions of secp256k1
operations. Once again, for x86 the idea is to support older non-BMI
machines, though at the cost of some performance versus using those
instructions when available (example on AMD EPYC):

  bignum_mul_p256k1               :    12.9 ns each (var  1.9%, corr -0.02) =   77632333 ops/sec
  bignum_mul_p256k1_alt           :    16.4 ns each (var  1.7%, corr -0.12) =   60796528 ops/sec

The ARM variants are optimizing for high-throughput multipliers and so
can be much faster there (example on Apple M1-based laptop):

  bignum_mul_p256k1               :    14.9 ns each (var  2.0%, corr 0.05) =   66971645 ops/sec
  bignum_mul_p256k1_alt           :     7.5 ns each (var  2.5%, corr 0.02) =  134015092 ops/sec

*Issue #, if available:*

*Description of changes:*


nebeid:
I suggest to add more documentation either in the .h include file or in each implementation a brief explanation about how this alt method is different from the non-alt, like the PR description so that this info stays within the code.

jargh:
OK, I have added some general explanation to the header file.

nebeid (approved):

PR #31, opened 2022-05-20
Symbol macros

justsmth:
*Description of changes:*

Change addresses two issues:
* Consuming projects previously had no easy way of hiding the s2n-bignum function/symbols in the shared libraries they produced. With this change when `S2N_BN_HIDE_SYMBOLS` is defined for the assembler,  the symbols will be marked hidden by an appropriate directive.
* Symbols were being exported in two forms, both with and without a leading underscore (`_`). The underscore was only needed on Apple platformes. The macros provided in this change produce only the appropriate symbols for the given platform.


justsmth:
The tests are passing on my x86 Linux host:
```
...
OK: word_recip(9236610159036633104) = 18393867477707217180
OK: word_recip(14560924854243060417) = 4922813171260366634
All OK
*** Partial tests (165) run; no failures but some skipped
*** x86 BMI and ADX instructions needed for other functions
```

I'll try to replicate the CI failures.

justsmth:
Will create another pull request after more testing is done.

PR #36, opened 2022-07-21
Basic point operations for the Weierstrass curves

jargh:
*Issue #, if available:*

*Description of changes:*


nebeid (approved):
It would be good if the PR description and/or the merge commit message contain a description of the gains and some benchmarking results, even if in range of percentages.

jargh:
These functions are functionally equivalent to the natural sequences of individual field operations. Efficiency of the point addition and mixed addition functions is not dramatically better than the corresponding sequence of field operations, typically around 5% depending on the platform (but it's also nice to have the whole function verified). The gains for the point doubling function are often more significant, typically in the range 15%-35% depending on curve and platform.

PR #38, opened 2022-10-28
Update s2n-bignum's open source license to include ISC

samuel40791765:
*Issue #, if available:*
`P73680511`

*Description of changes:*
AWS-LC used to be under the Apache2.0 license only. This creates potential licensing incompatibles with projects that have the GPLv2 open source license. AWS-LC has gone through its update, and this PR is to help s2n-bignum become truly compatible with AWS-LC and any other future customers (license-wise). More elaboration of why this is being done is included in the ticket. 

Although these are only text changes, this PR is massive, so let me know if I should split this PR up.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,9 @@
+SPDX-License-Identifier: Apache-2.0 OR ISC
+
+
+Apache 2.0 license
+-------------------------------------
+
 
                                  Apache License
                            Version 2.0, January 2004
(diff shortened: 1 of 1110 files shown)


samuel40791765:
Do we need to change the PR wording? It seemed fine to me, but I wanted to double check.
> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

hyandell (commented):
Source header changes are good.

LICENSE file seems confusing; we need to get that OR statement in there otherwise it looks confusingly like an AND. Simplest is to make that the first line in the LICENSE file. We could copy from the readme, but I think it'd be better to use SPDX here. So first line of LICENSE becomes:

SPDX-License-Identifier: Apache-2.0 OR ISC

PR wise, a good question. The existing language supports both licenses, it's basically ISC without the attribution requirement. In the future as the project matures we might want to move to DCO.

hyandell (approved):
LGTM :)

PR #41, opened 2023-01-06
Formalize UMOV of ARMv8

aqjune:
This PR formalizes [UMOV](https://developer.arm.com/documentation/ddi0596/2021-12/SIMD-FP-Instructions/UMOV--Unsigned-Move-vector-element-to-general-purpose-register-?lang=en) which is an alias of `mov` from an element of a NEON register to a general register:
```
mov <Rd>, <Rn>.<Ts>[<index>] (* aliases umov *)
```

Currently this PR supports `<Ts>` = `d` and `s`, but not `B` or `H` because they are not used in the vectorization pattern yet. Also it seems to add a slight complexity to the definition.
Also, this PR does not add `umov` as an alias of `mov` since `mov` was already (correctly) registered as an alias of `orr <reg>, zero`.

I tested the updated `decode` by running these two examples and checking the output:
```
(* mov x0, v0.d[1] , 4E183C00 *)
# let exp = `decode (word 0x4E183C00)` in DECODE_CONV exp;;
val it : thm = |- decode (word 1310211072) = SOME (arm_UMOV X0 Q0 1 8) 

(* mov w1, v2.s[3] , 0E1C3C41 *)
# let exp = `decode (word 0x0E1C3C41)` in DECODE_CONV exp;;
val it : thm = |- decode (word 236731457) = SOME (arm_UMOV W1 Q2 3 4)
```

I couldn't test the implemented semantics of umov which is `arm_UMOV` however. Before running random testing, I would like to write some simple unit tests that is analogous to the ones using `DECODE_CONV`. How can I write ones?

### DIFF
--- a/arm/proofs/decode.ml
+++ b/arm/proofs/decode.ml
@@ -11,6 +11,7 @@ let XREG_SP = new_definition `XREG_SP n = registers :> element n`;;
 let WREG_SP = new_definition `WREG_SP n = XREG_SP n :> zerotop_32`;;
 let XREG' = new_definition `XREG' (n:5 word) = XREG (val n)`;;
 let WREG' = new_definition `WREG' (n:5 word) = WREG (val n)`;;
+let QREG' = new_definition `QREG' (n:5 word) = QREG (val n)`;;
 
 let arm_logop = new_definition `arm_logop (opc:2 word) N
     (Rd:(armstate,N word)component) Rn Rm =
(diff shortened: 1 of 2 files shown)


jargh:
Looks good to me, at least from a human check against the ARM documentation. For more machine-checking, here's a pattern that could look at the semantics in more detail, setting up a program with just one instruction

let code_4E183C00 = define
 `code_4E183C00:byte list = [word 0x00; word 0x3C; word 0x18; word 0x4E]`;;

let execth = ARM_MK_EXEC_RULE code_4E183C00;;

then either checking concrete results like this:

let example = prove
 (`ensures arm
      (\s. aligned_bytes_loaded s (word pc) code_4E183C00 /\
           read PC s = word pc /\
           read Q0 s = word 12345678)
      (\s. read X0 s = word 0)
      (MAYCHANGE [PC; X0])`,
  ARM_SIM_TAC execth [1]);;

or simulating in a symbolic state so you can see the general pattern:
 
g `ensures arm
      (\s. aligned_bytes_loaded s (word pc) code_4E183C00 /\
           read PC s = word pc)
      (\s. read PC s = word(pc + 4))
      (MAYCHANGE something)`;;

e(ENSURES_INIT_TAC "s0" THEN ARM_VSTEPS_TAC execth [1]);;

aqjune:
Thanks, I confirmed that the concrete results of the two examples and one additional test are evaluated as expected. :)

PR #47, opened 2023-03-21
Switch to clean 8-bit array type in signature for high-level curve25519 algorithm APIs

torben-hansen:
**Issue #, if available:**

CryptoAlg-1601

**Description of changes:**

Switch API to use `uint8_t []` array type instead of `uint64_t []` array type. This is what applications typically operate with (in particular, AWS-LC).

It avoids a lot of copying around in AWS-LC which would otherwise add overhead. The copying is necessary because we can't cast array types, and interfaces are strictly checked in AWS-LC (even though it might just be decoration).

There are still a lot of interfaces to change, if we want to align everything to `uint8_t`!

The unit tests now throws interface type warnings. But maybe these can be solved in subsequent PRs. It would requires quite a few changes in `test.c`.

### DIFF
--- a/arm/curve25519/curve25519_ladderstep.S
+++ b/arm/curve25519/curve25519_ladderstep.S
@@ -5,7 +5,7 @@
 // Montgomery ladder step on pairs of (X,Z)-projective curve25519 points
 //
 // extern void curve25519_ladderstep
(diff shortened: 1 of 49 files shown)


torben-hansen:
Subset of https://github.com/awslabs/s2n-bignum/pull/48

PR #51, opened 2023-04-21
Add support for 8 more NEON instructions

aqjune:
*Description of changes:*

This pull requests adds supports for the following NEON instructions that are necessary to verify my `bignum_sqr_8_16`:

- and, add, dup, ext, movi, orr(mov), shrn, zip1

I attach a test .ml file that shows that (1) `decode` successfully decodes a few example instructions and (2) their symbolic executions show desirable output states:
https://gist.github.com/aqjune/aee276c8e4646dacad73f6a5575fa322

Also, I added a few comments that describe how functions in `PURE_DECODE_CONV` work.

I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


arm/proofs/decode.ml
```diff
@@ -592,6 +715,20 @@ let PURE_DECODE_CONV =
...
+     their values.
+     To understand this function further, you can start with the "LET"
+     case which is HOL Light's let binding `let x = e1 in e2` that
+     (1) evaluates e1, (2) takes `|- e1 = v1` and evaluates e2.
+     The benefit of using this continuation-passing style over repetitively
+     applying rewriting rules is its speed.
+     To understand the data structure of HOL Light's terms, you will want
+     to disable the term printer in OCaml REPL via
+     "#remove_printer pp_print_qterm;;". This can be enabled by
+     "#disable_printer pp_print_qterm;;" again. *)
```
aqjune, line 731: I left comments that explain functions based on my understanding. Please feel free to comment about the descriptions.

arm/proofs/instruction.ml
```diff
@@ -708,6 +708,26 @@ let arm_ADD = define
...
+            else simd16 word_add n m in
+          (Rd := d) s
+        else
+          let n:(64)word = word_subword n (0,64) in
+          let m:(64)word = word_subword m (0,64) in
+          let d:(64)word =
+            if esize = 32 then simd2 word_add n m
+            else if esize = 16 then simd4 word_add n m
+            else simd8 word_add n m in
+          (Rd := word_zx d:(128)word) s`;;
```
aqjune, line 729: I used `word_zx` instead of `word_subword` even if the latter can be also used to extend the bitwidth, because I thought readers might not be aware of the fact that `word_subword` can extend the bitwidth.
There is one inconvenience about `word_zx` which is that it is not supported by tactics like `WORD_BLAST`, but applying `REWRITE_TAC[word_zx]` before using `WORD_BLAST` seems to work.

arm/proofs/instruction.ml
```diff
@@ -1128,6 +1208,64 @@ let arm_UZIP1 = define
...
+    word_join (word_interleave2 xhi yhi) (word_interleave2 xlo ylo)`;;
+
+let word_interleave8 = new_definition
+ `(word_interleave8:(((N tybit0)tybit0)tybit0)word
+      ->(((N tybit0)tybit0)tybit0)word
+      ->((((N tybit0)tybit0)tybit0)tybit0)word)
+      x y =
+    let xlo,xhi = word_split_lohi x in
+    let ylo,yhi = word_split_lohi y in
+    word_join (word_interleave4 xhi yhi) (word_interleave4 xlo ylo)`;;
```
aqjune, line 1244: I am fine with moving these definitions to `Library/words.ml` if you think it is necessary.

arm/proofs/instruction.ml
jargh: Thanks for the observation! I believe I've now fixed the bug in WORD_BLAST (and related weaknesses around word_zx, word_sx) as a small addendum to this recent HOL Light update:

https://github.com/jrh13/hol-light/commit/6f6ac17e8f4bff6c7f6897de0570a004744e4963

jargh (approved):

PR #53, opened 2023-04-27
Adds script converting matched labels to per-file unique labels

torben-hansen:
*Description of changes:*

Script that was previously used to ensure labels are file-unique. Let's version it in s2n-bignum since it might be useful in the future.


jargh (approved):
Thanks, I've already used the script once by copying out of a previous PR.
So it clearly makes sense to have it in the repo.

PR #54, opened 2023-04-29
Support nonoverlapping reasoning of 128-bit loads and other access bitwidths

aqjune:
This patch adds support for nonoverlapping reasoning of 128-bit loads and other access bitwidths.
Also, it adds `bytes128` that is analogous to `bytes64`.
With this, the specification does not need to rely on the low-level `wbytes` definition anymore.

This patch allows s2n-bignum to prove this cute lemma which simply states that `stp` does not touch the values stored at `[x0]`:
https://gist.github.com/aqjune/3c72fbce77f04eefb97a83ca30cef361

Minor updates were necessary in a few proofs in x86/proofs/ because the changes in `READ_MEMORY_BYTESIZED_SPLIT` affected the proofs.

Also, this adds `.gitignore` that lists `.o`, `.obj`, `.correct` and `.a`. This hides the compiled binaries and previously proven facts from the list printed by `git status`.


aqjune:
I found that the CI failed, but it seems I don't have permission to look into the log.
Do you mind if I ask which proof is failing? Then, I will look into the failure.

aqjune:
It's fine. From my local EC2 instance I could reproduce that the following proofs failed:

```
ubuntu@ip-172-31-65-15:~/s2n-bignum/x86$ grep -i error */*.correct
p256/bignum_bigendian_4.correct:Error in included file /home/ubuntu/s2n-bignum/x86/proofs/bignum_bigendian_4.ml
p256/bignum_littleendian_4.correct:Error in included file /home/ubuntu/s2n-bignum/x86/proofs/bignum_littleendian_4.ml
p384/bignum_bigendian_6.correct:Error in included file /home/ubuntu/s2n-bignum/x86/proofs/bignum_bigendian_6.ml
p384/bignum_littleendian_6.correct:Error in included file /home/ubuntu/s2n-bignum/x86/proofs/bignum_littleendian_6.ml
p521/bignum_fromlebytes_p521.correct:Error in included file /home/ubuntu/s2n-bignum/x86/proofs/bignum_fromlebytes_p521.ml
```

... which seem to prove properties about pretty simple programs. :) Digging these further will be easier than investigating proofs of large arithmetic functions, I hope.

aqjune:
I found the reason and fixed the proofs so that they build successfully.

The reason was that the updated `READ_MEMORY_BYTESIZED_SPLIT` made `CONJUNCT1 READ_MEMORY_BYTESIZED_SPLIT` in the proofs pick up a wrong equality. They are fixed to use `CONJUNCT2` instead of `CONJUNCT1`.

Please feel free to merge this PR after #56 is merged first. :)

jargh (approved):
All looks very good now, thank you. Yes, I see a few pre-existing proofs were brittle with respect to the exact form of READ_MEMORY_BYTESIZED_SPLIT, so thanks for fixing those.

jargh:
I think it makes sense to merge this now. The other changes in #56 should be orthogonal anyway. Thanks!

PR #55, opened 2023-04-29
Add support for USRA (Vector)

aqjune:
This PR adds support for URSA (Vector).

Since its bit pattern shares the pattern of MOVI, they are combined into a single case and distinguished based on immh.

Test script: https://gist.github.com/aqjune/be15ad2d672515866ac772c6df3f47d2

### DIFF
--- a/arm/proofs/instruction.ml
+++ b/arm/proofs/instruction.ml
@@ -1187,6 +1187,31 @@ let arm_UMULH = define
         let d:N word = word((val n * val m) DIV (2 EXP dimindex(:N))) in
         (Rd := d) s`;;
 
+let arm_USRA_VEC = define
+ `arm_USRA_VEC Rd Rn shift esize datasize =
+    \s. let n:(128)word = read Rn s in
+        let n:(128)word =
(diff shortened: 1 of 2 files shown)


arm/proofs/instruction.ml
```diff
@@ -1187,6 +1187,31 @@ let arm_UMULH = define
...
+let arm_USRA_VEC = define
+ `arm_USRA_VEC Rd Rn shift esize datasize =
+    \s. let n:(128)word = read Rn s in
+        let n:(128)word =
+          if esize = 64 then usimd2 (\x. word_ushr x shift) n
+          else if esize = 32 then usimd4 (\x. word_ushr x shift) n
+          else if esize = 16 then usimd8 (\x. word_ushr x shift) n
+          else usimd16 (\x. word_ushr x shift) n in
+        let d:(128)word = read Rd s in
+        if datasize = 128 then
```
aqjune, line 1199: I think I can remove this if-else branch that is based on `datasize`; I will make this PR as ready after it is done.

PR #63, opened 2023-06-06
Add a simulator that tests the semantics of ARM instructions

aqjune:
This patch adds a simulator that tests the semantics of ARM instructions. The simulator was mostly written by John, and my only update is adding NEON instruction support and adding minor changes such as using random file names to avoid conflicts between two different simulator runs.

This

(1) Generates a random bitvector that can be decoded as an instruction, as well as random bitvectors for input values of register 

(2) Execute it with respect to s2n-bignum's operational semantics

(3) Compare the result with actual ARM CPU's output, by compiling a template code and run it

How to use: Open HOL Light at the s2n-bignum directory, then simply run `loadt "arm/proofs/simulator.ml";;`.
No need to load "arm/proofs/base.ml" because simulator.ml has `needs "arm/proofs/base.ml"`.

----

### DIFF
--- a/arm/proofs/simulator.ml
+++ b/arm/proofs/simulator.ml
@@ -1,0 +1,357 @@
+(* ------------------------------------------------------------------------- *)
+(* Encoding the registers and flags as a 32-element list of numbers.         *)
+(* ------------------------------------------------------------------------- *)
+
+needs "arm/proofs/base.ml";;
+
+let regfile = new_definition
+ `regfile s =
+   [val(read X0 s); val(read X1 s); val(read X2 s); val(read X3 s);
+    val(read X4 s); val(read X5 s); val(read X6 s); val(read X7 s);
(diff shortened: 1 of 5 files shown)


arm/proofs/simulator.ml
```diff
@@ -0,0 +1,358 @@
...
+  "xxx100101xxxxxxxxxxxxxxxxxxxxxxx";
+
+  (*** Extr ***)
+  "x00100111x0xxxxxxxxxxxxxxxxxxxxx";
+
+  (*** LSLV, LSRV ***)
+  "x0011010110xxxxx0010xxxxxxxxxxxx";
+
+  (*** UBFM ***)
+  (* TODO: This pattern can be extended to support SBFM after #61 is merged *)
```
aqjune, line 137: I left a comment regarding SBFM here :)

jargh (approved):
Very nice, and thanks for updating for SBFM!

PR #64, opened 2023-06-12
Update proofs CI to run the simulator

aqjune:
This patch updates the 'proofs' CI to run the simulator in parallel. It is set to 30 minutes by default.
Also, this adds a new Makefile target that runs the simulators. It can be run using `make sematest`.

To achieve parallelism, a custom bash script is written and invoked from Makefile instead of writing everything in Makefile.
This is because there is a challenge in relying on `make -j` to run simulators in parallel. `make -j` assigns one subprocess for one target object, but in this case, a simulator does not have any target object that it outputs. To resolve this, we can create N artifical outputs, but then another challenge is that in Makefile one cannot get the number of processes given to `-j`. :( (see also: https://stackoverflow.com/q/5303553/1488216)

`make sematest` always runs the simulators using all cores in CPU.

Also, additionally, `codebuild/tests.yml` is updated to build benchmark as well.

----

### DIFF
--- a/arm/Makefile
+++ b/arm/Makefile
@@ -369,3 +369,11 @@ sm2/%.correct: proofs/%.ml sm2/%.o ; (cd ..; (echo 'loadt "arm/proofs/base.ml";;
 run_proofs: $(PROOFS);
 
 proofs: run_proofs ; ../tools/check-proofs.sh .
+
+# Always run sematest regardless of dependency check
+FORCE: ;
+# Always use max. # of cores because in Makefile one cannot get the passed number of -j.
+# A portable way of getting the number of max. cores:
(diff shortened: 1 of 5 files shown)


aqjune:
This is the output of `make sematest` when it fails: https://gist.github.com/aqjune/cfe0d35640fa0bc0f9d5c0a92aa08f11 (simulator.ml was artificially modified to simulate a failure)
Output of `make sematest` when it succeeds: https://gist.github.com/aqjune/1a292f0cd60328f98c379b0670c84f4a

aqjune:
I will open this PR with aqjune-aws and see whether codebuild runs.

PR #65, opened 2023-06-13
Update sematests CI to run the ARM simulator

aqjune-aws:
This patch updates the 's2n-bignum-arm-sematests' CI to run the ARM simulator in parallel. Its running time is set to 30 minutes by default.

To achieve parallelism, a custom bash script is written and invoked from Makefile instead of writing everything in Makefile.
This is because there is a challenge in relying on `make -j` to run simulators in parallel. `make -j` assigns one subprocess for one target object, but in this case, a simulator does not have any target object that it outputs. To resolve this, we can create N artifical outputs, but then another challenge is that in Makefile one cannot get the number of processes given to `-j`. :( (see also: https://stackoverflow.com/q/5303553/1488216)

`make sematest` always runs the simulators using all cores in CPU.

Also, add '!/bin/bash' to the header of armsimulate.

This problem was found while Carl Kwan was running the simulator on his machine

----

### DIFF
--- a/arm/Makefile
+++ b/arm/Makefile
@@ -372,5 +372,9 @@ proofs: run_proofs ; ../tools/check-proofs.sh .
 
 # Always run sematest regardless of dependency check
 FORCE: ;
-# Dummy
-sematest: FORCE; echo "todo"
+# Always use max. # of cores because in Makefile one cannot get the passed number of -j.
+# A portable way of getting the number of max. cores:
+# https://stackoverflow.com/a/23569003/1488216
+NUM_CORES_FOR_SEMATEST = $(shell getconf _NPROCESSORS_ONLN)
+sematest: FORCE; ../tools/run-sematest.sh . $(NUM_CORES_FOR_SEMATEST) "$(HOLLIGHT)"
+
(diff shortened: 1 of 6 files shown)


aqjune-aws:
This is the output of make sematest when it fails: https://gist.github.com/aqjune/cfe0d35640fa0bc0f9d5c0a92aa08f11 (simulator.ml was artificially modified to simulate a failure)
Output of make sematest when it succeeds: https://gist.github.com/aqjune/1a292f0cd60328f98c379b0670c84f4a

aqjune-aws:
The s2n-bignum-arm-sematests CI has been fixed and is now working!

aqjune-aws:
From the CI log you can sometimes see the supposed-to-fail cases like this:
```
arm/proofs/armsimulate: line 17: 23060 Illegal instruction     (core dumped) $temp $*
random inst: decode 1919051390
OK: (fails correctly) instruction code 1919051390
```
1919051390 (`01110010011000100110011001111110`) is the ANDS instruction case with sf == 0 and N == 1 ([link](https://developer.arm.com/documentation/ddi0596/2021-12/Base-Instructions/ANDS--immediate---Bitwise-AND--immediate---setting-flags-)), which is UNDEFINED. Therefore the machine is trapped with Illegal instruction. Also s2n-bignum successfully fails decoding it. The `"x11100100xxxxxxxxxxxxxxxxxxxxxxx";` bitpattern in `iclasses` in the .ml file corresponds to this.

jargh (approved):

PR #70, opened 2023-07-04
Adding support for SHA256 and SHA512 intrinsics

kwancarl:
*Issue #, if available:*

*Description of changes:*

Adding new sha256.ml and sha512.ml files for definitions and conversions related to SHA256 and SHA512 intrinsics, respectively. Modified misc.ml in common/ and base.ml, decode.ml, instruction.ml, simulator.ml in arm/proofs/ to support simulation of SHA intrinsics on ARM model. 

### DIFF
--- a/common/misc.ml
+++ b/common/misc.ml
@@ -1081,6 +1081,12 @@ let (WORD_FORALL_OFFSET_TAC:int->tactic) =
 (* Do some limited simplification in association with symbolic execution.    *)
 (* ------------------------------------------------------------------------- *)
 
+(* Ocaml reference variable for platform specific conversions *)
+let extra_word_CONV = ref [NO_CONV];;
+
+(* Delay introduction of extra conversions *)
+let apply_extra_word_convs tm = FIRST_CONV (!extra_word_CONV) tm;;
+
 let ASSEMBLER_SIMPLIFY_TAC =
   let pth = prove
    (`!a b. a < a + bitval b <=> b`,
@@ -1096,6 +1102,7 @@ let ASSEMBLER_SIMPLIFY_TAC =
     DEPTH_CONV
      (GEN_REWRITE_CONV I [BITVAL_CLAUSES] ORELSEC
       WORD_RED_CONV ORELSEC
+      apply_extra_word_convs ORELSEC
       NORMALIZE_ADD_SUBTRACT_WORD_CONV ORELSEC
       NUM_RED_CONV ORELSEC
       INT_RED_CONV) THENC
--- a/arm/proofs/base.ml
+++ b/arm/proofs/base.ml
@@ -36,9 +36,35 @@ loadt "common/relational.ml";;
 loadt "common/interval.ml";;
 loadt "common/elf.ml";;
 
+(* ------------------------------------------------------------------------- *)
+(* Support for additional SHA instrinsics (from Carl Kwan)                   *)
+(* ------------------------------------------------------------------------- *)
+
+loadt "arm/proofs/sha256.ml";;
+loadt "arm/proofs/sha512.ml";;
+(* Adding extra conversions for SHA intrinsics *)
+extra_word_CONV :=
+        [SHA256H_REDUCE_CONV;
+         SHA256H2_REDUCE_CONV;
+         SHA256SU0_REDUCE_CONV;
+         SHA256SU1_REDUCE_CONV;
+         SHA512H_REDUCE_CONV;
+         SHA512H2_REDUCE_CONV;
+         SHA512SU0_REDUCE_CONV;
+         SHA512SU1_REDUCE_CONV]
+        @ (!extra_word_CONV);;
+
+(* ------------------------------------------------------------------------- *)
+(* The main ARM model.                                                       *)
+(* ------------------------------------------------------------------------- *)
+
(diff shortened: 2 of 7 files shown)


aqjune-aws:
FYI this is the tail of the log of the failed arm-sematests CI:
```
random inst: decode 3509383463
Stepping to state s1
CPU time (user): 0.322981
OK: arm_SUB X7 X9 (rvalue (word 2876))
random inst: decode 791615310
Stepping to state s1
CPU time (user): 0.82955
OK: arm_USRA_VEC Q14 Q26 17 32 64
arm/proofs/armsimulate: line 17: 11269 Illegal instruction     (core dumped) $temp $*
Error: term `3468722539`random inst: decode 3468722539
Simulator 1 failed!
```

Do you think 3468722539 is a SHA instruction?

aqjune-aws:
That being said, the cloud developer desktop seems to support Graviton 3 now. It provides machines starting with c7g: https://aws.amazon.com/ec2/instance-types/c7g/

I will investigate whether I can fix the CI use these.

common/misc.ml
```diff
@@ -1081,6 +1081,12 @@ let (WORD_FORALL_OFFSET_TAC:int->tactic) =
 (* Do some limited simplification in association with symbolic execution.    *)
 (* ------------------------------------------------------------------------- *)
 
+(* Ocaml reference variable for platform specific conversions *)
+let extra_word_CONV = ref [NO_CONV];;
+
+(* Delay introduction of extra conversions *)
+let apply_extra_word_convs tm = FIRST_CONV (!extra_word_CONV) tm;;
```
aqjune-aws, line 1088: This update actually looks nice - I should move SIMD-related rewriting rules to here as well if this patch is accepted.

aqjune-aws:
It seems switching to Graviton 3 is not straightforward (unless I am missing something) - it is this line (https://github.com/awslabs/s2n-bignum/blob/main/codebuild/cloudformation.yml#L177) that specifies s2n-bignum-arm-sematests must use an ARM instance, but I don't think we can specify its microarchitecture (the list of options is at https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html).
In the worst case we (1) should maintain a custom image based on Graviton 3, (2) Install QEMU or some emulator for this instruction. Or, we can simply negotiate and compare this with C-based implementation for CI...
I posted a question at a slack channel called "codebuild-interest" and am waiting for an answer.

aqjune-aws:
Ok, I got a reply that they only support Graviton 2 instances. :(
Which option should we pursue?

aqjune-aws:
The arm-sematests failure looks funny:
```
...
 68 [`read Q8 s1 = word (2 EXP 64)`]
 69 [`read PC s1 = word (pc + 4)`]

  `word (2 EXP 64) = word 18446744073709551616`
```
For some reason `read Q8 s1` couldn't be reduced to a natural number, remaining as an exponential expression. This made the equality check fail (even though they must be equivalent after the left hand side is reduced). :/

common/misc.ml
```diff
@@ -1096,6 +1102,7 @@ let ASSEMBLER_SIMPLIFY_TAC =
     DEPTH_CONV
      (GEN_REWRITE_CONV I [BITVAL_CLAUSES] ORELSEC
       WORD_RED_CONV ORELSEC
+      apply_extra_word_convs ORELSEC
```
aqjune-aws, line 1105: Oh - wait a second, since `apply_extra_word_convs` will effectively succeed everytime, and we are using ORELSEC here, does it imply that ` NORMALIZE_ADD_SUBTRACT_WORD_CONV ORELSEC NUM_RED_CONV ORELSEC INT_RED_CONV` won't run at all?

aqjune-aws: Nevermind, `FIRST_CONV` fails if all of the conversions fail. I thought that `FIRST_CONV` should always succeed.

jargh:
This is very nice to have, and so far I believe the modeling is fully
faithful and accurate, thanks!

But I think there are typos in the simulator bitmasks for a couple of
the SHA256 intrinsics, so they actually aren't being exercised by the
simulator when run. Looking at the ARM manual I think the appropriate
changes are as follows, but you'd better check.

```
   (*** SHA256H ***)
   "01011110001xxxxx010000xxxxxxxxxx"
->
   "01011110000xxxxx010000xxxxxxxxxx";

  (*** SHA256SU1 ***)
  "01011110000xxxxx110000xxxxxxxxxx";
->
  "01011110000xxxxx011000xxxxxxxxxx";

```
I don't deserve any prizes for being eagle-eyed, but I just looked
at the log of a long simulator run (with exactly the 8 SHA
instructions selected) and grepped to see if the balance looked about
right. If I change as above then it seems that it is, e.g.

```
$ grep arm_SHA LOG_SHA_INTRINSICS | grep OK | cut -f2 -d' ' | sort -n | uniq -c
     20 arm_SHA256H
     29 arm_SHA256H2
     18 arm_SHA256SU0
     23 arm_SHA256SU1
     22 arm_SHA512H
     16 arm_SHA512H2
     18 arm_SHA512SU0
     17 arm_SHA512SU1

```

jargh (approved):
All now looks good, thanks!

PR #71, opened 2023-07-23
Avoid using assembler macros and rename labels in the functions for RSA

aqjune-aws:
This patch updates the functions used by RSA signing to avoid using assembler macros because the `delocate` utility in BoringSSL cannot parse them. Also, the offsets in load/store instructions are simplified so that `delocate` can understand them as well.

Also, the labels having overlapping names are renamed so that aws-lc can compile them without duplicate symbol found errors.

---

### DIFF
--- a/arm/fastmul/bignum_emontredc_8n.S
+++ b/arm/fastmul/bignum_emontredc_8n.S
@@ -28,21 +28,20 @@
 // t,h should not overlap w,z
 // ---------------------------------------------------------------------------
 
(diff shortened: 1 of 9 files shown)


aqjune-aws:
I must confess that I haven't run proof checks before making this PR; I only ran tests after the changes. At least the tests did not report any error. Fingers crossed to the proof CIs.

jargh:
Looks good, all object files unchanged so as expected proofs all run.

jargh (approved):

PR #76, opened 2023-07-31
Make bignum_emontredc_8n's spec equal to its neon version

aqjune-aws:
This patch makes bignum_emontredc_8n's proof equal to its neon version, by moving `8 divides val k` out from the postcondition. Also, some assembly' specs are updated to use
`MAYCHANGE_REGS_AND_FLAGS_PERMITTED_BY_ABI`.

And this patch includes these minor non-functional changes as well:

- Multiplication functions' auxiliary lemma definitions are removed
- A copyright text is added to `simulator.ml` and `bignum_mul_8_16_neon.S`.

----

### DIFF
--- a/arm/fastmul/bignum_mul_8_16_neon.S
+++ b/arm/fastmul/bignum_mul_8_16_neon.S
@@ -1,3 +1,15 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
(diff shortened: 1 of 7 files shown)


aqjune-aws:
CI might fail because I haven't run full proof tests on my side; I will fix this PR if there is a failure

jargh (approved):
All looks good. It only makes sense to localize the preconditions in cases where those are really dependent on the input data (versus just the sizes).

PR #79, opened 2023-09-08
Add two ARM instructions and add print_log flags

aqjune-aws:
*Description of changes:*

This pull requests add support for the following two instructions:
- `str` for Q registers
- `bit z, x, y` which is conceptually `z[i] = y[i] ? x[i] : z[i]` for each bit.
For `str`, lemmas are updated to consider `bytes128`.
For `bit`, the simulator is updated to test it.

On top of these, this PR adds two flags for printing extra infos: `components_print_log` for `common/components.ml` and `arm_print_log` for `arm/proofs/arm.ml`. Once a flag is set, the tactics/conversions/functions within the corresponding file may print extra informations.
If `components_print_log` is set, `ASSUMPTION_STATE_UPDATE_TAC` prints assumptions that are erased due to any reason. This is helpful when a store instruction erases previous memory assumptions but cannot figure out why it happened.
If `arm_print_log` is set, `DISCARD_OLDSTATE_TAC` prints assumptions that are discarded because it uses not only 'old' states but also 'new' states. This is useful to understand why the result of load instruction is not reflected in the assumption.

---


aqjune-aws:
Hi - does this patch look fine?
I am revisioning my table lookup patch, and if this patch is merged it might be helpful in rebasing my next patch. :)

jargh (approved):
All looks good to me, thank you!

PR #81, opened 2023-09-12
Add bignum_copy_row_from_table and its Neon-variants for AArch64

aqjune-aws:
This patch adds constant-time table-lookup functions (`bignum_copy_row_from_table*`) and their proofs.
This patch only contains its AArch64 version, and the x86 version will follow later.
The failure of proving its x86 version seems to be related to handling negative offsets, and (if this is right) this can be avoided by simply proving positive offsets.
I will record this as a Github issue with a promise that the x86 scalar version will be provided after the RSA related things are finished.

This patch contains four table-lookup functions:
1. `bignum_copy_row_from_table`: a lookup for a generic table size
2. `bignum_copy_row_from_table_8n_neon`: a Neon version for a table whose width is a multiple of 8
3. `bignum_copy_row_from_table_16_neon`: Neon implementation of a table whose width is 16*64=1024 bits
4. `bignum_copy_row_from_table_32_neon`: Neon implementation of a table whose width is 32*64=2048 bits

The last two versions are initially written by Hanno Becker.

To successfully compile and run `test` and `benchmark` in x86, the scalar `bignum_copy_row_from_table` function is processed as a way similar to Neon functions.

The x86 version and its (ongoing) proof is stored in https://github.com/aqjune-aws/s2n-bignum/tree/tablelookup-x86 .
I will create a new github issue to record this after this is fully reviewed.

----

### DIFF
--- a/arm/Makefile
+++ b/arm/Makefile
@@ -154,6 +154,10 @@ BIGNUM_OBJ = curve25519/bignum_add_p25519.o \
              generic/bignum_cmul.o \
              generic/bignum_coprime.o \
              generic/bignum_copy.o \
+             generic/bignum_copy_row_from_table.o \
+             generic/bignum_copy_row_from_table_8n_neon.o \
+             generic/bignum_copy_row_from_table_16_neon.o \
+             generic/bignum_copy_row_from_table_32_neon.o \
              generic/bignum_ctd.o \
              generic/bignum_ctz.o \
              generic/bignum_demont.o \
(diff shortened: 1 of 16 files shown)


aqjune-aws:
I added a commit to update comments so that they say these are constant-time.

FWIW, merging this pull request will cause the git commit history to have these two individual commits.
If you want to have a single commit for each merged pull request in the git history, you can choose 'Squash and merge' from the dropdown menu (the down arrow in the 'Merge pull request' button), which will squash all commits into one. :)

jargh (approved):
This all looks really good, thanks!

PR #85, opened 2023-10-13
Let NONOVERLAPPING_TAC deal with more cases, improvements for speed 

aqjune-aws:
*Description of changes:*

This commit fixes `cmp` in `NONOVERLAPPING_TAC` so that it accepts more
expressions having operands of distinct constants, reducing a class of
possible failures from `NONOVERLAPPING_TAC`.

Also, this commit performs various optimizations to reduce the latency
of running proofs:

- Unuse `VALID` in `NONOVERLAPPING_TAC` because `VALID` is slow
- Fix `DISCARD_ASSUMPTIONS_TAC` to have linear complexity
- Add a cache to `ORTHOGONAL_COMPONENTS_CONV` so that it does not run
  `prove()` for a once-proven goal. Also, make it fail early when the goal
  is not `orthogonal_components _ _`; this is important in Arm because
  `ARM_CONV` invokes `COMPONENT_READ_OVER_WRITE_CONV` which uses
  `ORTHOGONAL_COMPONENTS_CONV` this may traverse
  non-`orthogonal_components` terms.
- Add a cache to `ASSIGNS_SWAP_CONV`.

---

### DIFF
--- a/arm/proofs/arm.ml
+++ b/arm/proofs/arm.ml
@@ -394,18 +394,6 @@ let ARM_VERBOSE_SUBSTEP_TAC exths sname g =
 (* Throw away assumptions according to patterns.                             *)
 (* ------------------------------------------------------------------------- *)
 
-let DISCARD_ASSUMPTIONS_TAC P =
-  REPEAT(FIRST_X_ASSUM(K ALL_TAC o check P));;
-
-let DISCARD_MATCHING_ASSUMPTIONS pats =
-  DISCARD_ASSUMPTIONS_TAC
-   (fun th -> exists (fun ptm -> can (term_match [] ptm) (concl th)) pats);;
-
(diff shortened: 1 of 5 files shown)


aqjune-aws:
Oops, I pushed a wrong code. Rerunning...

aqjune-aws:
I added a new commit that fixes the quadratic behavior of `DISCARD_ASSUMPTIONS_TAC`.

aqjune-aws:
This excel compares the running times (in sec.) of proofs before/after this patch.
[stats.xlsx](https://github.com/awslabs/s2n-bignum/files/13051953/stats.xlsx)

jargh (approved):
This is very much faster with many assumptions, as well as cleaner. Thank you!

PR #87, opened 2023-10-28
Ed25519 support and related updates

jargh:
*Issue #, if available:*

*Description of changes:*


x86/proofs/edwards25519_encode.ml
```diff
@@ -0,0 +1,152 @@
...
+      ==> ensures x86
+           (\s. bytes_loaded s (word pc) windows_edwards25519_encode_mc /\
+                read RIP s = word pc /\
+                read RSP s = stackpointer /\
+                read (memory :> bytes64 stackpointer) s = returnaddress /\
+                WINDOWS_C_ARGUMENTS [z; p] s /\
+                bignum_pair_from_memory(p,4) s = (x,y))
+           (\s. read RIP s = returnaddress /\
+                read RSP s = word_add stackpointer (word 8) /\
+                read RSP s = word_add stackpointer (word 8) /\
```
aqjune-aws, line 143: Nit: two `read RSP`s here..!

arm/proofs/bignum_mod_m25519_4.ml
```diff
@@ -0,0 +1,123 @@
...
+  CONJ_TAC THENL [REAL_INTEGER_TAC; ALL_TAC] THEN
+
+  ACCUMULATOR_POP_ASSUM_LIST(MP_TAC o end_itlist CONJ o DESUM_RULE) THEN
+  REWRITE_TAC[REAL_BITVAL_NOT; m_25519] THEN
+  DISCH_THEN(fun th -> REWRITE_TAC[th]) THEN
+  CONV_TAC(RAND_CONV REAL_POLY_CONV) THEN REAL_INTEGER_TAC);;
+
+let BIGNUM_MOD_M25519_4_SUBROUTINE_CORRECT = time prove
+ (`!z x n pc returnaddress.
+      nonoverlapping (word pc,0x58) (z,8 * 4)
```
aqjune-aws, line 111: Would the specification be more readable if we use `length bignum_mod_m25519_4_mc` instead of constant `0x58`? This could be applied to other files.

aqjune-aws:
As a starting point, I checked that the 15 new functions of PR invoked from `benchmark.c` and `test.c`.
Also, their specifications (`*_SUBROUTINE_CORRECT`) appeared in the successfully verified specification list in the arm-proofs and x86-proofs CI. (For x86-proofs ci, the new 30 specifications appeared. It is 15*2 = 30 because we have WINDOWS versions too)
Also, I checked these things:
- The `_alt` function and their non-alt variants have equivalent specifications (modulo must-be-different items like program size)
- For each function, its Arm spec vs. x86 spec vs. x86 (windows) spec are equivalent.
- The functions do not write to the stack memory with variable size.

What is left is briefly checking whether the test + specifications look okay..? I believe this needs review from people who have knowledge in elliptic curves. (demonstration of successful integration of this patch w.r.t. latest AWS-LC might be great witness for the validity of specifications too)
The `test.c` contains invocation of these new functions and comparison of their output w.r.t. the reference outputs.
The assembly files - specifications pairs are collected from this PR and listed in this snippet:
https://gist.github.com/aqjune-aws/853888d22eeaed3500207e6af9b75ef6
- `!a b c` is for-all values a, b and c.
- `nonoverlapping (p,n) (q,m)` means that two memory ranges [p,p+n) and [q,q+m) do not overlap.
- `ensures arm P Q R` means that for any input state satisfying `P`, after running this program, an output state satisfying `Q` will be reached, `R` is the relation between input state and output state that it needs to satisfy.
- `aligned_bytes_loaded s (word pc) bignum_invsqrt_p25519_mc` means that at location `pc`, a machine code (which is a byte list) `bignum_invsqrt_p25519_mc` is loaded.

x86/proofs/edwards25519_encode.ml
jargh: Thank you for catching that, will fix.

arm/proofs/bignum_mod_m25519_4.ml
jargh: Yes, this is certainly true; it might well make sense to write all the "code size" specs like this, which would be more readable and less brittle (though a bit more verbose, particularly in code + data cases with data tables included in the size). I will not change it right here, but we should perhaps add this to the to-do list of general improvements to s2n-bignum.

jargh: Thank you for the careful general check and extracting the specifications! This is probably exactly the right thing for people who are not concerned with the inner workings of the HOL Light proofs to look at in order to decide whether the functionality is right.

jargh: I've fixed the issue June noted above, as well as switching edwards25519_scalarmul* functions to the divstep-based modular inverse in line with the curve25519 scalar multiplications.

torben-hansen (approved):

PR #88, opened 2023-11-01
Document that x25519 function does not implement zero-check

torben-hansen:
*Description of changes:*

Section 6.1 of [rfc7748](https://www.rfc-editor.org/rfc/rfc7748#section-6.1) specifies that x25519 implementations MAY perform a zero-check to eliminate a risk; avoid bad actor from voiding a peer contribution to shared secret.

Simply document that this is **not** performed by this implementation (which could really be deduced by `void` return type. But just make it explicit in function documentations.


jargh (approved):
Must be a good thing to document this precisely.

If we later decide that the zero check is so useful we would actually like to return a flag with the function, that would also be pretty easy.

PR #89, opened 2023-11-01
Use vfree_in instead of free_in if valid

aqjune-aws:
This patch uses `vfree_in` instead of `free_in` if valid.
Actually, for the .ml files in `{arm,x86}/proofs` directory all occurrences of `free_in` could be replaced with `vfree_in` because their first operand was a variable.

This does not shorten the running time of CI, but for some proofs small speedups ~5% have been observed.

----


jargh (approved):
Definitely a worthwhile optimization.

PR #91, opened 2023-11-13
Update x86 simulator.ml to add the latest cases

aqjune-aws:
*Description of changes:*

This patch updates x86's simulator.ml to contain the latest output of the script after the pull request for ed25519 is merged.
The script is written above the definition of `iclasses` in the file, and small edits were done by hand on its output to properly wrap them with brackets. :)

----


jargh (approved):
Thanks, it's good to keep things in step to make sure we are potentially exercising all the iforms that are used in the code.

Issue #97, opened 2023-11-21
[portability] Do not use static qualifiers for Windows build


torben-hansen:
Using static qualifiers inside arrays, [as done in the s2n-bignum header file](https://github.com/awslabs/s2n-bignum/blob/main/include/s2n-bignum.h), can aid the compiler and is specified in C99 6.7.5.3/7. But it's not portable, since mscv apparently doesn't support it. Consider whether it's worth using this feature.

This means that downstream must maintain their header files.

torben-hansen:
See e.g. https://developercommunity.visualstudio.com/t/c11-function-declarators-not-supported/10097089

jimgrundy:
How do we feel about #ifdefs? These kinds of qualifiers are not just meaningful to compilers, they give users an understanding of the the behavior of a routine with a degree of precision that a comment doesn't quite match.

torben-hansen:
That would work. But you would either have to macrofy things or duplicate declarations. Or make it generative, reading assembly files, from one source of truth.

jimgrundy:
Part of me would be happy, well, less unhappy than simply not having it, to do something as repetitive as the following (though I can also see the objection to it):

#ifdef MSVC
extern void bignum_add_p25519 (uint64_t z[/*static 4*/], uint64_t x[/*static 4*/], uint64_t y[/*static 4*/]);
#else 
extern void bignum_add_p25519 (uint64_t z[static 4], uint64_t x[static 4], uint64_t y[static 4]);
#endif

But ideally we would generate that from a single source of truth. Maybe also generate Rust extern signatures from that same source of truth one day.

jimgrundy:
I guess I could live with just dropping the "static 4" in the interim since the generation solution seems like the ultimate way to go.

torben-hansen:
Maybe something stupid like this can work: https://github.com/awslabs/s2n-bignum/pull/98

aqjune-aws:
Closing this issue as resolved by the pull request.

PR #98, opened 2023-11-22
Replace static array qualifier with macro that maps to nothing if it detects an msvc compiler

torben-hansen:
*Issue #, if available:*

https://github.com/awslabs/s2n-bignum/issues/97

*Description of changes:*

MSVC doesn't support static qualifiers in arrays. Improve portability, without any ugly duplication, by using a macro. The macro will map to nothing if an MSVC compiler is detected...

Similar issues with pre C99 and C++; https://wiki.sei.cmu.edu/confluence/display/c/API05-C.+Use+conformant+array+parameters.

### DIFF
--- a/include/s2n-bignum.h
+++ b/include/s2n-bignum.h
@@ -24,33 +24,40 @@
 //        The "_neon" forms target machines with NEON instructions.
 // ----------------------------------------------------------------------------
 
+
+#if defined(_MSC_VER) || !defined(__STDC_VERSION__) || __STDC_VERSION__ < 199901L || defined(__STDC_NO_VLA__)
+#define S2N_BIGNUM_STATIC
+#else
+#define S2N_BIGNUM_STATIC static
+#endif
+
 // Add, z := x + y
 // Inputs x[m], y[n]; outputs function return (carry-out) and z[p]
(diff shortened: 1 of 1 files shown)


aqjune-aws:
The static is useful for catching possible out of bounds accesses, esp., when using clang. Therefore I believe it is a good idea to enable it whenever it is possible, as this PR does.
For example, clang is silent with this code: https://godbolt.org/z/9aqz6GccY
But with the static modifier, clang emits a nice warning message: https://godbolt.org/z/njzY8d4z8

Regarding the macro condition, I found this interesting article that has its own macro condition at "Exceptions" section:
https://wiki.sei.cmu.edu/confluence/display/c/API05-C.+Use+conformant+array+parameters
Should we leave a comment mentioning that checking the MSVC_VER macro is a crude way of doing this?

torben-hansen:
MSVC always emits `_MSC_VER`. So, I used that https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#Internal_version_numbering. But there are other known sharp edges. Capturing that as well now.

aqjune-aws (approved):
Since (1) the macro condition successfully captures the MSVC compiler (https://godbolt.org/z/9Eh55adKh) (2) I inspected the changes and looked okay (3) CI is happy with this, I will accept this patch and merge.

Issue #82, opened 2023-09-17
x86 version of constant-time table lookup is missing in #81


aqjune-aws:
#81 only delivers constant-time table lookup for Arm, but not x86.
The x86 version and its (ongoing) proof is stored in https://github.com/aqjune-aws/s2n-bignum/tree/tablelookup-x86 . I will revisit this after RSA-related things are done and finish its proof.
The failure that I've met while proving its x86 version seems to be related to handling negative offsets, and (if this is right) this can be avoided by simply proving positive offsets. I created an issue that is assigned to myself not to forget.

PR #100, opened 2023-12-27
Add verified bignum_copy_row_from_table for x86

aqjune-aws:
This patch adds a constant-time table-lookup function for x86 and its proof. This resolves issue #82.
Also, it contains the following changes:
- The specification of Arm bignum_copy_row_from_table is slightly updated to use `LENGTH bignum_copy_row_from_table_mc` rather than its actual number.
- `x86_print_log` flag is added, to do the things what `arm_print_log` does.
- Let `NONOVERLAPPING_TAC` print the goal when it fails, to help users.

Resolving the origin of the problem (dealing with negative offset in nonoverlapping reasoning) was simply avoided by rewriting the relevant loop assembly code from decreasing counter form to increasing counter form. I think this issue must be addressed at some point. For constant-time table lookup, this wasn't a critical problem however. I have an idea for reducing the time that proof CI takes, and if it is successful, the time necesary to address this kind of problem will be significantly reduced... I think this is the right order to tackle.

*Issue #, if available:*

#82

*Description of changes:*

### DIFF
--- a/arm/proofs/bignum_copy_row_from_table.ml
+++ b/arm/proofs/bignum_copy_row_from_table.ml
@@ -114,9 +114,12 @@ let READ_MEMORY_BYTES_BYTES64 = prove(`!z s.
               GSYM BIGNUM_FROM_MEMORY_BYTES; BIGNUM_FROM_MEMORY_SING]);;
 
 
-let BIGNUM_COPY_ROW_FROM_TABLE_SUBROUTINE_CORRECT = prove(`!z table height width idx pc n m returnaddress.
-    nonoverlapping (word pc, 0x6c) (z, 8 * val width) /\
-    nonoverlapping (word pc, 0x6c) (table, 8 * val height * val width) /\
+let BIGNUM_COPY_ROW_FROM_TABLE_SUBROUTINE_CORRECT = prove(
+  `!z table height width idx pc n m returnaddress.
+    nonoverlapping (word pc, LENGTH bignum_copy_row_from_table_mc)
+                   (z, 8 * val width) /\
+    nonoverlapping (word pc, LENGTH bignum_copy_row_from_table_mc)
+                   (table, 8 * val height * val width) /\
     nonoverlapping (z, 8 * val width) (table, 8 * val height * val width) /\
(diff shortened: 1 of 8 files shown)


jrh13:
Looks really good overall, thank you! Actually I had not realized that cmov was guaranteed to read a memory operand. That's useful to know for other situations. Just one small suggestion:

When I try it either on x86 and ARM versions of MacOS (of course I can't run it on the latter but it is intended to build), the assembler from the clang toolchain doesn't seem happy with the movq line, so maybe this is more portable?

-        movq    [rax], 0
+        mov     QWORD PTR [rax], 0

Also, the idea of printing the failed goal in NONOVERLAPPING_TAC is a good one, but I haven't actually checked that it doesn't lead to a large amount of "spam" if the tactic is speculatively applied in situations where it may fail. Do you know if this happens? I could just look at the logs of a build. If that is a problem it may make sense to have it selectable.

jrh13:
A victim of markup: I meant that instead of the first line use the second
```
-        movq    [rax], 0
+        mov     QWORD PTR [rax], 0
```

aqjune-aws:
> When I try it either on x86 and ARM versions of MacOS (of course I can't run it on the latter but it is intended to build), the assembler from the clang toolchain doesn't seem happy with the movq line, so maybe this is more portable?

I updated the line to use mov QWORD PTR. :)

> Also, the idea of printing the failed goal in NONOVERLAPPING_TAC is a good one, but I haven't actually checked that it doesn't lead to a large amount of "spam" if the tactic is speculatively applied in situations where it may fail. Do you know if this happens? I could just look at the logs of a build. If that is a problem it may make sense to have it selectable.

The log will only appear when `components_print_log := true` is set (which is a flag that has been added in the past) or `{X86,ARM}_STEP_TAC` fails (which won't happen in successful proofs). :) It won't appear when running proofs already in s2n-bignum.

jrh13:
OK, this now looks great! And thanks for the reassurance about NONOVERLAPPING_TAC. I will approve.

jrh13 (approved):

jrh13 (approved):

jargh (approved):

PR #101, opened 2024-01-04
Print running times of proofs, factor out proof runner

aqjune-aws:
This patch makes `*.correct` files contain running time in seconds as well as its start and end time.
This is to find out which proofs take most of the time from CI. According to the CPU usage graph, there are only two or three proofs that use most of the CI running time, and they even seem to start late during parallel runs. If they start early, we can reduce significant amount of CI time!
As a part of this update, this patch factors out proof runner script commands from inline commands in Makefile as a separate shell script file.

*Issue #, if available:*

*Description of changes:*


aqjune-aws:
Confirmed that the time stamps are now being printed at CI log & the original proof check functionally is unaffected. Merging...

aqjune (approved):

aqjune-aws:
Closing this and will open with aqjune to approve using aqjune-aws

aqjune-aws:
CI is rerunning for some reason... but they have passed before creating the new PR. :/

PR #104, opened 2024-01-22
Allow MIT-0 license as well as Apache-2.0 and ISC

jargh:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -204,7 +204,7 @@ OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 MIT-0 license
 -------------------------------------
 
-Copyright Amazon.com, Inc. or its affiliates.
+Copyright 2021-2024 Amazon.com, Inc. or its affiliates.
 
 Permission is hereby granted, free of charge, to any person obtaining a
 copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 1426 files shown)


LICENSE
```diff
@@ -200,3 +199,24 @@ ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
 WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
 ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
 OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+
+MIT-0 license
+-------------------------------------
+
+Copyright Amazon.com, Inc. or its affiliates.
```
aqjune-aws, line 207: I looked at https://github.com/aws/mit-0 and it seems we need a year - do you want to add it?

aqjune-aws (approved):

LICENSE
jargh: Yes, a good point. Do you know if it should be the current year (which in a sense is when it was first MIT-0 licensed) or the year range since inception of the project? That is, whether it is appropriate to write  "2024" or "2021-2024"?

PR #105, opened 2024-01-31
Fix decoding of MOVSXD in the x86 model

jargh:
This corrects the order of sizes in the non-trivial 32->64 case, as well as properly distinguishing and handling the (admittedly degenerate) 32->32 case. The other degenerate case 16->16 is rejected by design, as with most other instructions using operand size override prefixes.

This also adds new cases to the simulator to validate MOVZX and MOVSX/MOVSXD, which are not currently used in s2n-bignum code and hence were not previously cosimulated against hardware in standard checks.

*Issue #, if available:*

*Description of changes:*


aqjune-aws (approved):
I did the two following checks and they were okay.

(1) I compared the results of `DECODE_CONV [...]` with x86 online decoder ([this](https://defuse.ca/online-x86-assembler.htm#disassembly2)) for each item in the new byte lists added to `x86/proofs/simulator.ml`. They were consistent with what the online decoder said.

(2) I ran x86 simulator on my laptop, after leaving the newly added byte lists only. Running it for 10 minutes did not catch any problem, so.... I think this patch is okay!

jargh:
Thanks! I didn't know that online decoder - a nice resource,

I'll wait to confirm all checks pass then merge,

PR #108, opened 2024-02-09
Lenngren-based X25519 for non-alt ARM code

jargh:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/arm/curve25519/curve25519_x25519.S
+++ b/arm/curve25519/curve25519_x25519.S
@@ -7,7 +7,8 @@
 //      https://github.com/Emill/X25519-AArch64/blob/master/X25519_AArch64.pdf
 //      https://github.com/Emill/X25519-AArch64
 //
-// and the SLOTHY-based re-engineering of that code by Hanno Becker:
+// and the SLOTHY-based re-engineering of that code by Abdulrahman, Becker,
+// Kannwischer and Klein:
 //
 //      https://eprint.iacr.org/2022/1303.pdf
 //      https://github.com/slothy-optimizer/slothy/tree/main/paper
@@ -202,9 +203,8 @@ curve25519_x25519_scalarloop:
 // In particular, the basic dataflow and the organization between integer
 // and SIMD units is identical, with only a few minor changes to some
 // individual instructions (for miscellaneous reasons). The scheduling
-// was redone from scratch by SLOTHY starting from Hanno Becker's
-// un-interleaved form and using the same scripts as in Becker et al's
-// paper.
+// was redone from scratch by SLOTHY starting from the un-interleaved
+// form in the SLOTHY work cited above, and using the same scripts.
 //
 // The intermediate value annotations were added to provide data that
 // is used in the formal proof, indicating which lines assign specific
--- a/arm/curve25519/curve25519_x25519.S
+++ b/arm/curve25519/curve25519_x25519.S
@@ -1,6 +1,18 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT-0
 
+// **********************************************************************
+// This code is substantially derived from Emil Lenngren's implementation
+//
+//      https://github.com/Emill/X25519-AArch64/blob/master/X25519_AArch64.pdf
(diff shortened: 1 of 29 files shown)


aqjune-aws:
Hi, since the updated functions now use Neon, should we update test/benchmark.c to run them only when neon is enabled?

Or should we simply assume that Arm always has Neon? I am not sure whether AWS-LC can rely on this assumption, though.

Emill:
Reference: https://developer.arm.com/documentation/102525/0100/Compiling-for-Neon-with-Arm-Compiler-6:

> Neon is required in all standard Armv8-A implementations, so targeting any Armv8-A architecture or processor will allow the generation of Neon code.

Not sure about Armv9-A though.

aqjune-aws:
I checked these things:
- The updates to the main specifications `CURVE25519_X25519_SUBROUTINE_CORRECT` and `CURVE25519_X25519_BYTE_SUBROUTINE_CORRECT` contained minor changes such as used stack size/code size/modifiable register set, and I thought these were okay
- The update to the precise definition of callee-save registers regarding Q8~Q15 (thanks.)
- The theorems printed in the proofs CI are okay (did not change compared to the master branch)
- The update to the postcondition of `BIGNUM_INV_P25519_SUBROUTINE_CORRECT` which looks correct to me.
```
(\s. read PC s = returnaddress /\
                  (coprime(p_25519,n)
                   ==> bignum_from_memory(z,4) s = inverse_mod p_25519 n))
```
=>
```
(\s. read PC s = returnaddress /\
                  bignum_from_memory(z,4) s =
                  (if p_25519 divides n then 0 else inverse_mod p_25519 n))
```

I will create a follow-up pull request that removes the existing Neon check condition from test.c/benchmark.c .

I think the lemmas `SIMD_SPLIT_JOIN_CLAUSES`, `EXTRA_SPLIT_JOIN_CLAUSES` can probably be used in other places using SIMD registers in s2n-bignum as well.

aqjune-aws (approved):

aqjune-aws:
If this is merged, AWS-LC might want to link these assembly files only when Neon is supported. In theory this might be not necessary, but wanted to let people know :)

arm/curve25519/curve25519_x25519.S
```diff
@@ -1,6 +1,18 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT-0
 
+// **********************************************************************
+// This code is substantially derived from Emil Lenngren's implementation
+//
+//      https://github.com/Emill/X25519-AArch64/blob/master/X25519_AArch64.pdf
+//      https://github.com/Emill/X25519-AArch64
+//
+// and the SLOTHY-based re-engineering of that code by Hanno Becker:
```
hanno-becker, line 10: Can we change this to "Abdulrahman, Becker, Kannwischer, Klein"?

arm/curve25519/curve25519_x25519.S
```diff
@@ -26,841 +38,1317 @@
...
+        ldr     x0, [swap]
+        cmp     x0, x2
+        str     x2, [swap]
+
+// The following inner loop code is derived closely following Lenngren's
+// implementation available at "https://github.com/Emill/X25519-AArch64".
+// In particular, the basic dataflow and the organization between integer
+// and SIMD units is identical, with only a few minor changes to some
+// individual instructions (for miscellaneous reasons). The scheduling
+// was redone from scratch by SLOTHY starting from Hanno Becker's
```
hanno-becker, line 205: Can we say "... starting from the un-interleaved developed by Becker et. al in the SLOTHY paper", or just "... developed in the SLOTHY paper [ref]"?

hanno-becker:
This is amazing, thank you @jargh.

arm/curve25519/curve25519_x25519.S
jargh: Thank you June for the additional followup work and the careful review!

I will fix those attributions as Hanno suggests now before we merge.

jargh:
I've tweaked the attributions, which will trigger a re-run of all checks.
Assuming nothing untoward happens, I'll merge after that.

PR #109, opened 2024-02-11
Enable testing bignum_copy_row_from_table on x86

aqjune-aws:
I would like to have this reviewed after #108 is merged to avoid possible merge conflicts.

This is a small patch that enables testing bignum_copy_row_from_table on the x86 architecture.

This had to be done in the past when the x86 implementation was being added, but was missing.

*Issue #, if available:*

*Description of changes:*


jargh (approved):
This is definitely a good idea. I have tried the modified code and it works as expected, performing the benchmarking also on x86.

PR #113, opened 2024-03-06
Define program equivalence and rewrite `bignum_{mul,sqr}_8_16_neon` proofs using it

aqjune-aws:
This patch rewrites the proofs of correctness of bignum_mul_8_16_neon and bignum_sqr_8_16_neon using program equivalence. Their `*_SUBROUTINE_CORRECT` theorems (which are the most important theorems) remain unchanged. Their proofs now rely on the program equivalence between their scalar versions and Neon versions.

To define program equivalence and prove related lemmas, this patch adds a bunch of definitions and lemmas in the following new files:
- `common/relational2.ml`: This file defines relational Hoare logic for machine code as a new predicate `ensures2`. This internally uses a new predicate called `eventually_n`. This file also defines a unary Hoare logic predicate `ensures_n` which uses `eventually_n` in contrast to `eventually` which is used by `ensures`. Finally, this file proves several properties on `ensures2`, `ensures_n` and `ensures`.
- `arm/proofs/equiv.ml`: This file includes simulation tactics for a pair of programs. Also, this includes tactics and lemmas that enable conversion between Arm `ensures_n` and `ensures` because this is important to 'induce' the final specification from equivalence and original specification
- `tools/diff.py`: This is a Python script that prints the diff lines between two assembly files as an OCaml list of tuples. This is in turn used by `EQUIV_STEP_TAC`.

This patch only targets use cases in Arm, but `common/relational2.ml` is designed to be reusable for x86 assembly programs as well.

Also, this patch adds a new powerful function `needs_skip_proofs`. This is equivalent to 'needs', but skips checking all proofs using CHEAT_TAC. This must be used very carefully. I left a comment suggesting that this must be only called at the beginning of REPL (or a file) only when you are sure that all theorems in the loaded file were proven to be correct in the past.

*Issue #, if available:*

*Description of changes:*


aqjune-aws:
The failure is because `Toploop.use_input` which is internally used by `needs_skip_proofs` does not exist in OCaml 4.05.
What do you think about upgrading CI to use OCaml 4.14?

jargh:
This is now in good shape for merging, thanks.

I do still have just one tiny suggestion, which is to
remove the commented-out proof on lines 767-783
of "arm/proofs/bignum_sqr_8_16_neon.ml". Since
there is now a real proof, this might just be
confusing, and possibly a bit frightening if you
just crudely grep the sources :-)

To return to the big picture, this is really a very
significant methodological improvement, thank you!

aqjune-aws:
Oops, I thought I removed the comment but it was still there. I guess I removed it from somewhere else...
I removed the lines.

jargh (approved):
This is now all great, thank you! I'll merge...

PR #114, opened 2024-03-15
Make `make proofs` fail early, use hol.sh

aqjune-aws:
This patch
- Fixes `make proof` to fail early.
- Uses the new `hol.sh`.

This patch does not upgrade the OCaml version of s2n-bignum CI to 4.14 yet.

--


jargh (approved):
This looks a nice improvement, and it all works for me, thanks!

PR #116, opened 2024-03-23
Check axioms at the end of run-proof.sh

aqjune-aws:
This patch adds `check_axioms()` and calls it at the end of `run-proofs.sh` to check the absence of newly introduced axioms.

This also adds grepping 'exception' from the output of REPL because `failwith` appears as `Exception: Failure ...`.


jargh (approved):
Nice to have these additional checks, thanks!

PR #119, opened 2024-04-04
Update Codebuild to switch to Zarith, use num instead of Int

aqjune-aws:
This patch replaces `Int` with `num` because `Int` does not exist anymore.
A relevant HOL Light patch is: https://github.com/jrh13/hol-light/pull/94


aqjune-aws:
Arm proofs: 4 hr 47 min -> 4 hr 37 min
x86 proofs: almost no change (4 hr)
It became 10 minutes faster. :)

jargh (approved):
All looks good, thanks!

PR #120, opened 2024-04-11
Maintain a list of specification theorems and check against it

aqjune-aws:
The current `make proofs` checks that all specifications are proven by comparing the number of appearances of `let [A-Z_0-9]*_SUBROUTINE_CORRECT` in the `proofs/*.ml` files with respect to `SUBROUTINE_CORRECT : thm` in `*.correct` output files.

It seems this can be fixed to be more robust, by addressing these concerns:

- Simply comparing the numbers is not as strong as comparing the exact list of specifications (between those to be proven and are actually proven)
- If a patch is large, an author might mistakenly remove/rename a specification and a reviewer might miss it as well. However, the current checker does not raise an alarm in this case.

This patch adds two text files that contain explicit lists of specifications that must be proven for each architecture.
The list for {Arm,x86} is located at `{arm,x86}/proofs/specifications.txt`. These can be generated by `tools/collect-specs.sh {arm,x86} > {arm,x86}/proofs/specifications.txt`.

An author has to update these files if new specifications are added to s2n-bignum. On the other hand, on the reviewers' side, looking into these files is enough to confirm that the reviewing patch adds/removes specifications. :)

*Issue #, if available:*

*Description of changes:*


jargh (approved):
Yes, this is a nice improvement in reliability of checking, despite the small extra work per new function. Thanks!

PR #124, opened 2024-05-13
Mention which tool s2n-bignum relies on to check proofs

aqjune-aws:
*Description of changes:*

This pull request briefly adds to README.md that to check proofs in s2n-bignum HOL Light must be installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,8 +314,13 @@ models make some simplifications and idealizations but model pretty faithfully
 the way in which specific machine instructions modify registers, flags and
 memory.
 
-To perform the formal proof for a particular function, just use the Makefile
-within either the `arm` or `x86` directories to generate a target of the form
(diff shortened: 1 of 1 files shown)


aqjune-aws:
CI checks are failing because (1) hol.sh is updated to pick `ledit` as a line editor, and (2) it cannot find a shared library for 'Unix' (which is used to measure running times). The reason why it cannot find a shared library for Unix module is not clear.
My two cents is that using `make switch` will resolve these two problems. For (1), ledit will be installed using `make switch`. For (2), if the failure was due to a change in the virtual machine setting, using make switch will create a consistent environment.
#125 will address this.

aqjune-aws:
CI checks are fixed by #125 separately.

jargh (approved):
Makes good sense, thanks.

PR #125, opened 2024-05-13
Use make switch to set environment

aqjune-aws:
*Issue #, if available:*

*Description of changes:*


aqjune-aws:
CI checks are successfully fixed to use `make switch`.

jargh:
This seems to work very nicely - I tried it out manually on a machine where the native OCaml is different from 4.14, and both ARM and x86 proofs work smoothly with the localized 4.14 installation.

jargh (approved):

PR #126, opened 2024-05-27
Add `needs "{arm,x86}/proofs/base.ml"` to proofs

aqjune-aws:
This patch adds `needs "{arm,x86}/proofs/base.ml";;` to the beginning of every proof file, instead of externally running this from `run-proof.sh`.

This has several benefits:

1. This makes it clear that proofs require loading `"{arm,x86}/proofs/base.ml"` in advance, which was not explicitly described in README.

2. When using OCaml REPL, this removes waiting finishing `loadt "{arm,x86}/proofs/base.ml";;`
For example, to run `arm/proofs/bignum_add.ml` on OCaml REPL, `loadt "arm/proofs/base.ml";;` must be run first, which will take a few minutes. This patch removes the necessity of intervention.

3. vscode-hol-light's release candidate 2.0 now supports 'Go to definition'. This requires the dependency between source files to be explicitly written as `needs`,`loadt`,`loads`.
This patch makes the functionality work well for s2n-bignum proofs.

### How to review this patch

This patch can be reviewed as follows:
- `{arm/x86}/proofs/specifications.txt` are not modified by this patch. CI checks guarantee that the theorems are not deleted.
- Diff of this patch only contains `needs "{arm,x86}/proofs/base.ml";;\n` (not other texts) in every .ml file (https://dellsystem.me/posts/additions-deletions-git). This guarantees that the specification theorems are not modified.
- `tools/run-proof.sh` does not load the `base.ml` file anymore.

*Issue #, if available:*

### DIFF
--- a/arm/proofs/bignum_add.ml
+++ b/arm/proofs/bignum_add.ml
@@ -7,6 +7,8 @@
 (* Addition of bignums.                                                      *)
 (* ========================================================================= *)
 
+needs "arm/proofs/base.ml";;
+
 (**** print_literal_from_elf "arm/generic/bignum_add.o";;
  ****)
 
--- a/arm/proofs/bignum_add_p25519.ml
+++ b/arm/proofs/bignum_add_p25519.ml
@@ -7,6 +7,8 @@
 (* Addition modulo p_25519, the field characteristic for curve25519.         *)
 (* ========================================================================= *)
 
+needs "arm/proofs/base.ml";;
+
 (**** print_literal_from_elf "arm/curve25519/bignum_add_p25519.o";;
  ****)
 
--- a/arm/proofs/bignum_add_p256.ml
+++ b/arm/proofs/bignum_add_p256.ml
@@ -7,6 +7,8 @@
 (* Addition modulo p_256, the field characteristic for the NIST P-256 curve. *)
 (* ========================================================================= *)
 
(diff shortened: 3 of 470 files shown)


aqjune-aws:
The CI failures are from the camlp5 version issue. https://github.com/jrh13/hol-light/pull/99 will resolve this problem.

jargh:
I like this idea on all the grounds mentioned above, but maybe it would be tidier to avoid duplicating the "needs" line in cases where it was already present (which seems to be mainly, but not exclusively, the full-point operations). And, by the way, this makes "consistency" another argument for the change since there is no real logic behind which files previously happened to have the line.

If I do:
``
for i in */proofs/*.ml; do echo -n "$i "; grep needs $i | grep base | wc -l; done | sort -k 2 -n
``
it shows up 122 such cases.

aqjune-aws:
I fixed this patch to avoid introducing needs base.ml if there already exists 'needs "..../base.ml";;'. Now running your bash script prints entries with count 1 only. I used this script: https://gist.github.com/aqjune-aws/f9e95a4ca0cfd817b7c09ca702274d99
Some files are edited in this patch as well to have consistent indentation at the beginning of the file. :)

jargh (approved):
Looks perfect now, thanks!

PR #127, opened 2024-06-12
Remove linear scan of decoded instruction in symbolic execution tactics 

aqjune-aws:
*Description of changes:*

To find the decoded instruction at the current PC,
the symbolic step tactics (`{X86,ARM}_STEP_TAC`) are performing linear scan of
a long conjunction containing `{arm,x86}_decode` predicate at each location.
The long conjunction is a part of output of `{X86,ARM}_EXEC_RULE` which
returns `(length of mc is blah) /\ (decoded instructions are blah)`.

This patch is reducing the linear time complexity to constant time
by rather constructing the result of decoding as an array from constant PC offset
to a theorem stating the result of decoding, e.g.,:
```
|- !s pc.
       bytes_loaded s (word pc) (BUTLAST bignum_mux_mc)
       ==> x86_decode s (word (pc + 5)) (3,XOR (% r9) (% r9));
```
To achieve this, in this patch, `{X86,ARM}_EXEC_RULE` now returns `thm * (thm option array)` where
where the first `thm` of the pair is `LENGTH .._mc = ..` and the
second item is the array from PC offset to the decoded instruction theorem.
This reduces `{X86,ARM}_STEP_TAC` (precisely speaking `{X86,ARM}_THM` in them)
the previous linear time complexity with respect to the number of total instructions in a program
to constant time.

Luckily, the necessary changes in the proofs was pretty small and systematic.
The necessary changes were from something like
```
REWRITE_(TAC or RULE) [...; *_EXEC; ..]
```
to
```
REWRITE_(TAC or RULE) [...; fst *_EXEC; ..]
```
because this rewrites were only using the `LENGTH .._mc = ..` part.

Another improvement that is orthogonal to the above topic is to do `CLARIFY_TAC` after
each step of equivalence tactic. This also makes the speed faster.

---


jargh (approved):
Thank you, this is a good idea to take the linear-time lookup out of the core instruction simulation, and I like the way it's done :-)

PR #131, opened 2024-07-10
P-256 scalar multiplication and related tweaks

jargh:
*Issue #, if available:*

*Description of changes:*


aqjune-aws:
Hello all,

I will(can) review this patch whether this patch contains necessary updates to tests, benchmarks, etc, but reviewing the more important part (the functional correctness) needs help from people with cryptography background since my view can be inaccurate.

Reviewing the functional correctness can begin with finding out the new or updated `*_SUBROUTINE_CORRECT` variables from diffs. For example, p256's scalar multiplication for x86 is newly added in this patch and its specification is described in the new variable which looks like this:
```
let P256_SCALARMUL_SUBROUTINE_CORRECT = time prove
 (`!res scalar point n xy pc stackpointer returnaddress.
        ALL (nonoverlapping (word_sub stackpointer (word 1368),1368))
            [(word pc,0x84fb); (scalar,32); (point,64)] /\
        ALL (nonoverlapping (res,64))
            [(word pc,0x84fb); (word_sub stackpointer (word 1368),1376)]
        ==> ensures x86
             (\s. bytes_loaded s (word pc) p256_scalarmul_mc /\
                  read RIP s = word pc /\
                  read RSP s = stackpointer /\
                  read (memory :> bytes64 stackpointer) s = returnaddress /\
                  C_ARGUMENTS [res;scalar;point] s /\
                  bignum_from_memory (scalar,4) s = n /\
                  bignum_pair_from_memory (point,4) s = xy)
             (\s. read RIP s = returnaddress /\
                  read RSP s = word_add stackpointer (word 8) /\
                  !P. P IN group_carrier p256_group /\
                      affinepoint_p256 xy P
                      ==> affinepointz_p256
                            (bignum_pair_from_memory(res,4) s)
                            (group_pow p256_group P n))
          (MAYCHANGE [RSP] ,, MAYCHANGE_REGS_AND_FLAGS_PERMITTED_BY_ABI ,,
           MAYCHANGE[memory :> bytes(res,64);
                     memory :> bytes(word_sub stackpointer (word 1368),1368)])`,
```
I think the specific parts that needs help are numerical definitions. For the above example, in a precondition it says `bignum_from_memory (scalar,4) s = n /\ bignum_pair_from_memory (point,4) s = xy` stating that in a buffer at pointer `scalar` a number `n` is stored and its bitwidth is 64 * 4 = 256 bits, and in a buffer at pointer `point` two 256-bit numbers are consecutively stored and its 512-bit representation is `xy`. In its postcondition, there is `!P. (P IN group_carrier p256_group /\ affinepoint_p256 xy P) ==> affinepointz_p256 (bignum_pair_from_memory(res,4) s) (group_pow p256_group P n)` (with parentheses explicitly added) saying that if we interpret the number `xy` as an affine point P (I am understanding 'affine point' it as just the canonical (x,y) 2-dimensional point on an elliptic curve) and P is on the elliptic curve of P256, then the resulting point is P^n. My above interpretation might not make sense with respect to what crypto people typically think about what EC point operations are; it will be greatly helpful if the new/updated specifications can be compared with what crypto peoples' understandings.

aqjune-aws (approved):
Tested test/benchmark and checked that equivalent specifications (non-alt vs. alt, x86 vs arm) are consistently updated.
Regarding the details of functional correctness, discussed internally and confirmed that they are ok.

PR #133, opened 2024-07-12
Avoid using term_match for performance, do small refactoring to equiv tactics

aqjune-aws:
This patch makes small updates to avoid term_match which seems to be slower than directly using the match statement. A particular case that had to be fixed was when term_match was being invoked for every assumption because the number of assumptions can grow linearly to the number of simulated instructions unless it is cleaned up in the middle. This fix specifically brought ~14% speedups to proofs using equivalence checking tactic because when the lockstep simulation simulates one program it had to filter assumptions to hide that correspond to another program and it was using term_match.
There was also some speedups in large proofs not using equivalence tactic. I will attach the full result in the PR.

Another update that was orthogonally make was to the code related to equivalence checks :

- Rename `mk_equiv_statement` to `mk_equiv_statement_simple` that only considers a pair of straight-line codes from begin to end, and add `mk_equiv_statement` for more generic case.
- Some fix in `common/relational2.ml` in order to support constant-time proof in the future.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/arm/proofs/arm.ml
+++ b/arm/proofs/arm.ml
@@ -380,11 +380,8 @@ let ARM_THM =
       let pc_expr = fst (List.hd inst) in
       if is_var pc_expr then 0
       else try
-        let _,inst2,_ = term_match [] `pc_base + ofs` pc_expr in
-        let ofs,ofs_var = List.hd inst2 in
-        if ofs_var <> `ofs:num` || not (is_numeral ofs)
-        then failwith ""
-        else dest_small_numeral ofs
+        let pc_base,ofs = dest_binary "+" pc_expr in
+        dest_small_numeral ofs
       with Failure _ ->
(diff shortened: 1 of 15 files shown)


aqjune-aws:
[table.xlsx](https://github.com/user-attachments/files/16195890/table.xlsx)

aqjune-aws:
After the scalar multiplication code is merged, the CI check time of x86 proofs (according to CodeBuild log) is shortened by 10 min, and Arm proofs is shorted by 1 second. Before the scalar multiplication was added, Arm proof was shortened by 15 min and x86 proof was by 10 minutes. Therefore the 1 second in Arm can be simply due to the last long-running proof not taking advantage from this patch.

jargh (approved):
All looks good, a surprisingly large performance gain!

PR #135, opened 2024-07-25
P-256 precomputed point scalar multiplication

jargh:
*Issue #, if available:*

*Description of changes:*


include/s2n-bignum.h
```diff
@@ -995,6 +995,11 @@ extern void p256_montjmixadd_alt(uint64_t p3[S2N_BIGNUM_STATIC 12],uint64_t p1[S
 extern void p256_scalarmul(uint64_t res[S2N_BIGNUM_STATIC 8],uint64_t scalar[S2N_BIGNUM_STATIC 4],uint64_t point[S2N_BIGNUM_STATIC 8]);
 extern void p256_scalarmul_alt(uint64_t res[S2N_BIGNUM_STATIC 8],uint64_t scalar[S2N_BIGNUM_STATIC 4],uint64_t point[S2N_BIGNUM_STATIC 8]);
 
+// Scalar multiplication for precomputed point on NIST curve P-256
+// Input scalar[4], blocksize, table[]; output res[8]
+extern void p256_scalarmulbase(uint64_t res[S2N_BIGNUM_STATIC 8],uint64_t scalar[S2N_BIGNUM_STATIC 4],uint64_t blocksize,uint64_t *table);
```
rod-chapman, line 1000: Can table == NULL? If not, how many uint64_t's is it supposed to be pointing at? Initialized to 0, uninitialized, or don't care? Proper precondition here please!

include/s2n-bignum.h
```diff
@@ -995,6 +995,11 @@ extern void p256_montjmixadd_alt(uint64_t p3[S2N_BIGNUM_STATIC 12],uint64_t p1[S
...
+extern void p256_scalarmulbase_alt(uint64_t res[S2N_BIGNUM_STATIC 8],uint64_t scalar[S2N_BIGNUM_STATIC 4],uint64_t blocksize,uint64_t *table);
+
```
rod-chapman, line 1002: As above

include/s2n-bignum.h
nebeid: Where can I find the explanation of the format of the table, i.e., the number of precomputed points, how they're precomputed, how to tradeoff?

jargh: There is a somewhat more detailed explanation in the comment banner at the top of the code, although it probably needs expanding in more detail:
`
Given scalar = n and point = P, assumed to be on the NIST elliptic
curve P-256, the input argument "table" is expected to be a table of
multiples of the point P in Montgomery-affine form, with each block
corresponding to "blocksize" bits of the scalar as follows, where
B = 2^{blocksize-1} (e.g. B = 8 for blocksize = 4):
For each i,j with blocksize * i <= 256 and 1 <= j <= B
the multiple 2^{blocksize * i} * j * P is stored at
tab[8 * (B * i + (j - 1))], considered as uint64_t pointers
or tab + 64 * (B * i + (j - 1)) as byte pointers.
`

rod-chapman: So... it's a constant read-only look-up table. OK so far... but the _user_ is supposed to construct this (how???) and supply it?  What happens if it is incorrect? This API seems to fail the "easy to use correctly, hard to use incorrectly" giggle test.

jargh: The best thing is probably for s2n-bignum to provide an additional function to construct the table given the parameter "blocksize". This could either be used dynamically as part of a general initialization step for AWS-LC or other customer library, or just called offline to generate a static table. Having an s2n-bignum function also keeps things more abstract/modular: if we later wanted to change something like the style of recoding, we could change both in step. Does that seem to make sense?

Such a function is easy to adapt from the code already used inside the `tests/test.c` testing file, namely the function `reference_multable`. So maybe a verified version of that function (replacing the reference versions with s2n-bignum P-256 functions)? I would still leave the user to allocate the memory when calling it (otherwise we have to start worrying about the semantics of calling `malloc`), but that's a fairly easy calculation.

rod-chapman: But seriously... why are we providing so much choice? How many different tables might the user want to generate? I would rather just pre-generate them and give the user 3 simple calls calls for "fast", "faster" and "fastest" with stated time/space tradeoffs...

and yes... no malloc() please! Surely the table goes on the stack, assuming you know exactly how big it is for some given set of parameters. Or pre-generated tables, which are static and invisible to the user.

nebeid: I'll let dkostic chime in on this. Currently, the precomputed tables for the base points are part of the read-only code. They are generated by a Go script.

jargh: In reply to Rod, I'd say that since the three versions proposed are so similar, and almost nothing is lost performance-wise by the parametrization, there's a lot to be said for just one piece of code - s2n-bignum is already rich enough in bifurcations (or multifurcations, if that's a word) of versions.

rod-chapman: I agree - give the user fewer choices, and zero ways to make a mess. Can the table(s) be auto-generated directly into a .S file and then just compiled in to the library as a static/global constant?

aqjune-aws (approved):
I checked that test/benchmark executed successfully on x86/arm and and the new specifications looked fine w.r.t. each other (alt vs. non-alt, x86 vs arm, scalarmulbase vs non-base).
If adding more table is a concern, what about adding these as a separate pull request? This will help avoiding a larger pull request and easier review. :)

include/s2n-bignum.h
jargh: Indeed, it's actually easier in proof terms to generate specific validated tables than to verify the function to generate them. (There are already verifications of specific tables in some of the Edwards functions where these are hard-coded into the function, and the proofs are pretty easy.) There might however be a need to tweak the ELF handling to handle object files that are just a data section (say a .rodata section). That's something independently useful anyway.

ctz: I'm a little late to this conversation, but from my side the flexibility is good. eg, at the moment in code that uses the p256 field operations here , I have a larger constant table of multiples of the curve generator point, plus a smaller runtime-generated table of multiples of a given public key point.

PR #137, opened 2024-07-26
Add hybrid `p256_montjadd` and `p256_montjdouble` for Arm, slow multipliers

aqjune-aws:
This patch adds a hybrid (scalar + SIMD) version of `p256_montjadd` and `p256_montjdouble` for Arm with slow multipliers.
This uses the hybrid field operations (`bignum_montmul_p256_neon`, `bignum_montsqr_p256_neon`).

Original:
```
p256_montjadd                   :   513.2 ns each (var  0.1%, corr  0.04) =    1948395 ops/sec
p256_montjdouble                :   276.3 ns each (var  0.1%, corr  0.01) =    3619774 ops/sec
```
New (this patch): both are 11% faster
```
p256_montjadd                   :   461.4 ns each (var  0.1%, corr  0.13) =    2167340 ops/sec
p256_montjdouble                :   248.4 ns each (var  0.3%, corr  0.06) =    4025645 ops/sec
```
To prove this, a clean call-based version is written and placed at `arm/p256/unopt` . The assembly is compiled only when `make proofs` is run, and not included in the library.
The optimized version is placed in `arm/p256` and it is the result of applying function inlining + eliminating redundant load/store instructions + folding (add addr, const) + load/store.

Orthogonal changes are:
- `common/actions_merger.ml` adds OCaml functions that can be used in proofs to combine two lists of actions that are created from, say, diff(fileA,fileB) and diff(fileA,fileC). The combined action corrsponds to diff(fileA,fileC).
- Added `extra_early_rewrite_rules` which has a set of rewrite rules to apply at the early stage of `ASSEMBLER_SIMPLIFY_TAC`. This is helpful when one wants to break memory read of 128 bit into `word_join` of two 64 bit reads.
- `tools/diff.py` now does not use the Python `difftools` package because the package sometimes generate suboptimal diffs. Instead, it invokes the Linux `diff` commandline tool and uses its standard output. Also, its name is changed to `gen-actions.py`.
- Added a cache to `COMPONENT_CANON_CONV` to avoid repetitive computation. This brings ~2% speedup to proofs
- Many other improvements and equivalence checking tactics


aqjune-aws:
Hoping that CI check is happy this time...

The equiv. check between unoptimized p256_montjadd and optimized p256_montjadd takes slightly less than 2 hours. This is longer than 1 hr + alpha that I introduced before.
The reason is like this. For faster eq. checking, I initially added a custom orthogonal_compnents cache to `ORTHOGONAL_COMPONENTS_CONV` in order to avoid reproving `orthogonal_components` between one memory access and one register. This was used by p256_montjadd's equiv checker to speed up symbolic simulation further. However, I removed this cache later because this solution looked a bit hacky. It seems that more fundamental problem is that a set of MAYCHANGE of p256_montjadd grows big during simulation because the number of words accessed in memory are accessed. Once it grows big, a single instruction can take up to 1 second which is bad. :( I think a better solution is to merge memory locations that are adjacent in MAYCHANGE so that the number of entries is kept small.

aqjune-aws:
CI check: proof for p256_montjadd passes, the remaining failure is from scalar multiplication proof (specifically `P256_SCALARMUL_SUBROUTINE_CORRECT`). Fixing this doesn't seem hard, but since #135 is also adding base point scalar multiplication it seems to be a good idea to wait until #135 is merged.

aqjune-aws:
Updated p256_scalarmul to use the optimized point ops & field ops.
TODO is to update p256_scalarmulbase to use the optimized field ops.

jargh (approved):
All great :-) I really like the new organization of using the unopt forms in the proof. In the future we can probably make a lot of the high-level proofs simpler and cleaner following this methodology.

PR #138, opened 2024-08-05
Add global assumptions paragraph

aqjune-aws:
This patch adds a short paragraph about the assumptions that s2n-bignum globally assume, on top of the precondition of the formal specifications.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,6 +407,28 @@ impressionistic terms we can turn it into English as follows:
 
     - The machine flags (also freely modifiable according to the ABI)
 
+**Global Assumptions.**
+In addition to the assumptions described in the formal specifications,
+s2n-bignum implementations globally assume that the execution environment is
+configured as following:
+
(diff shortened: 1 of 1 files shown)


aqjune-aws:
Added that code and stack pointer alignments are explicitly mentioned in the formal specs.

jargh:
I like this clarifying material. A couple of suggestions: 

* Perhaps we should refine the little-endian part to be a bit more nuanced: the code (probably) works big-endian but we don't prove it rigorously. Something like: "we believe all code works equally well on a big-endian machine, but we do not validate that fact ourselves, and the instruction model underlying the formal proof does not directly address this question since it is assuming little-endian" 
* In the case of x86 (maybe ARM too?) should we say explicitly that we assume 64-bit mode?

jargh (approved):
Indeed, it's maybe a bit redundant, but nice. All good now, thanks!

Issue #136, opened 2024-07-25
x86 test CI check should confirm consistency between x86_att and x86


aqjune-aws:
A note from jargh:
```
It could basically just do a "make clobber; make" inside x86_att and check whether any of the generated code changed (either via git status or making initial copies and comparing or something)
```

Issue #134, opened 2024-07-15
Check that Arm cosimulation encompasses all Arm instructions appearing in implementations


aqjune-aws:
The current version of Arm cosimulator does not check that it encompasses all Arm instructions appearing in implementations.
For x86 cosimulator, this issue does not exist because it gathers all x86 instructions from .ml files.

PR #139, opened 2024-08-10
Update Arm cosimulator to check that opcodes appear, add x86_att test to CI

aqjune-aws:
This patch resolves these following issues:

- #134: Check that Arm cosimulation encompasses all Arm instructions appearing in implementations
- #136: Now CI checks consistency between x86 and x86_att

Also, this patch makes a small update in run-sematest.sh to not check exit code but grab failure/exception. This is because exceptions may not always be connected to non-zero exit, and also ledit does not propagate exit code (see https://github.com/chetmurthy/ledit/issues/4).

*Issue #, if available:*

*Description of changes:*


jargh (approved):
These are both very valuable improvements in our automated checking, thank you!

PR #142, opened 2024-08-21
Adopt the Arm SIMD-optimized p384 fields to point operations

aqjune-aws:
This patch updates the p384 point operations to use the field operations optimized using NEON and the SLOTHY optimizer.

The performance improvement is around 9%.

```
p384_montjdouble                :   591.7 ns each (var  0.1%, corr -0.02) =    1690105 ops/sec
p384_montjadd                   :  1143.6 ns each (var  0.1%, corr  0.04) =     874447 ops/sec
p384_montjscalarmul             : 329982.3 ns each (var  0.0%, corr  0.09) =       3030 ops/sec
=>
p384_montjdouble                :   543.2 ns each (var  0.1%, corr  0.02) =    1840798 ops/sec
p384_montjadd                   :  1044.5 ns each (var  0.1%, corr -0.09) =     957396 ops/sec
p384_montjscalarmul             : 303017.8 ns each (var  0.0%, corr  0.04) =       3300 ops/sec
```

This patch also includes the following updates:
- Add `arm/proofs/utils` and factor out the OCaml codes that are parameters to the equivalence checking tactics as files in the directory.
- Update EQUIV_STEP_TAC to take additional arguments that describe dead value information of the registers in the programs. The information describes which registers contain dead values after each program location. EQUIV_STEP_TAC uses this information to clean assumptions that will not be used later during simulation.
- A few tactics for equivalence checking to shorten repeatedly appearing proof patterns and a few updates in arm.ml for better error messages
- Bug fixes in actions_merger.ml
- Add a user-defined custom cache to ORTHOGONAL_COMPONENTS_CONV. This is useful for equiv checking because the memory invariants that appear in assumptions have a typical form (the byte64).
- And many other improvements for speed
*Issue #, if available:*

*Description of changes:*


arm/proofs/arm.ml
```diff
@@ -922,14 +946,21 @@ let ARM_ADD_RETURN_STACK_TAC =
    `(!x. (!y. P x y) ==> (!y. Q x y)) ==> (!x y. P x y) ==> (!x y. Q x y)`
   and sp_tm = `SP` and x30_tm = `X30`
   and dqd_thm = WORD_BLAST `(word_zx:int128->int64)(word_zx(x:int64)) = x` in
-  fun execth coreth reglist stackoff ->
+  fun ?(pre_post_nsteps:(int*int) option) execth coreth reglist stackoff ->
```
aqjune-aws, line 949: I had to add this optional argument to `ARM_ADD_RETURN_STACK_TAC` because the previous heuristics for determining the number of instructions used to do callee-save register save/restore did not match the cases in p384_montjadd/double.S . Since this is an optional argument, the existing proofs simply sets it to None and hence has no effect.

arm/proofs/arm.ml
```diff
@@ -740,10 +756,13 @@ let ARM_MACRO_SIM_ABBREV_TAC =
     fun n ->  if n = 0 then pat0
               else vsubst[mk_small_numeral n,pan] patn
   and grab_dest =
-    let pat = `read (memory :> bytes(p,8 * n))` in
+    let pat = `read (memory :> bytes(p,8 * n)) s` in
```
aqjune-aws, line 759: This part was updated because sometimes `ARM_MACRO_SIM_ABBREV_TAC` was grabbing the wrong `read (memory :> bytes ...)` expression from the postcondition and deciding that was the output buffer, such as `read (memory :> bytes ...) s_input` (where `s_input` is some imaginary input state). I resolved this solution by explicitly stating that it must use a variable `s:armstate` where `s` is the variable provided by the lambda function of the post condition. Also updated at line 742 that `s` must be the postcondition's state variable.

aqjune-aws:
CI checks succeeded (7.7 hrs for p384 montjscalarmul)! Trying to make it faster...

jargh (approved):
Thank you for the nice performance improvement :-) It all looks good to me, so  approving and merging now.

PR #147, opened 2024-09-22
Toplevel P-521 operations

jargh:
*Issue #, if available:*

*Description of changes:*


aqjune-aws:
- test and benchmark passed for the affected files
- Specs of the new functions are added to specifications.txt and they are equivalent to their p384 versions modulo the p256/p384 constant uses and code sizes
- modular_decode is not using the first element of the input pair in its definition

aqjune-aws (approved):

PR #150, opened 2024-10-07
Tactic optimization: hoist term/type parsings, remove unnecessary term_matchs, etc

aqjune-aws:
*Description of changes:*

This patch hoists term/type parsings and remove unnecessary term_matchs and unused definitions.
Also, this makes equivalence checker print infos only when `arm_print_log` is set.
With these simple updates combined, some proofs have slightly less than a 10% speedup.

[speedup.xlsx](https://github.com/user-attachments/files/17272280/speedup.xlsx)

By submitting 
this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


jargh (approved):
This is great, thank you! It's always a good idea to get as much partial evaluation done as possible, and it's nice that it gives such good speedups.

Issue #143, opened 2024-08-21
Check the usage of Arm x18 in continuous integration checks


aqjune-aws:
x18 should not be used because it is not a plainly usable register in some environments.

PR #156, opened 2024-11-05
Check usage of the x18 register in Arm

aqjune-aws:
This patch adds checking the usage of x18 register in Arm. According to
https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms , x18 should not be used for Apple platforms.
Resolves: https://github.com/awslabs/s2n-bignum/issues/143

This also removes old comments about installing hol-light using apt.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/arm/Makefile
+++ b/arm/Makefile
@@ -50,14 +50,18 @@ SPLIT=tr ';' '\n'
 
 ifeq ($(ARCHTYPE_RESULT),aarch64)
 ASSEMBLE=as
+OBJDUMP=objdump -d
 else
 ifeq ($(ARCHTYPE_RESULT),arm64)
 ASSEMBLE=as
+OBJDUMP=objdump -d
 else
 ifeq ($(OSTYPE_RESULT),Darwin)
 ASSEMBLE=as -arch arm64
+OBJDUMP=otool -tvV
 else
 ASSEMBLE=aarch64-linux-gnu-as
+OBJDUMP=aarch64-linux-gnu-objdump -d
 endif
 endif
 endif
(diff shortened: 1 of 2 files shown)


aqjune-aws:
The script was slightly updated to remove lines containing `.quad` before checking because some version of objdump interprets this as instructions (and it contained x18).

jargh (commented):
I've tested this by deliberately switching a register to x18 and indeed it triggered a build failure with an informative message, so the overall structure seems good.

But I found that on my current Mac OS X (ARM), it still seems necessary to have the explicit stdin indicator "-" in the assembler line, including the first one where the x18 checking gets inserted. I tried adding the same final "-" as in the last line, and then it seems to work fine.
````
cat curve25519/curve25519_x25519_byte.S | sed -e 's/\/\/.*//' | cc -E -I../include  -xassembler-with-cpp - | tr ';' '\n' | grep -v -E '^\s+.quad\s+0x[0-9a-f]+$' | as -o curve25519/curve25519_x25519_byte.o

clang: error: no input files
make: *** [curve25519/curve25519_x25519_byte.o] Error 1

aqjune-aws:
Thanks. I did not run the latest version on Mac OS X ... I fixed it at the last commit, and also added another minor update.

jargh:
Thanks, let's hope all the CI works smoothly now. I have also tested it on an older x86 Mac and it worked well.

jargh (approved):
Thanks, now seems to work fine on all platform combinations I've tried (Mac OS X and Linux / ARM and several different x86). In all cases I tested that the basic build works, and also that it correctly fails if I artificially use x18 somewhere.

PR #157, opened 2024-11-06
Fix HOL Light commit hash to explicitly state which version was used

aqjune-aws:
This patch updates the CI script to explicitly check out a specified HOL Light version.
This helps users understand which version of HOL Light is being used for CI. Previously it was simply picking the latest version of HOL Light at the CI running time.

*Issue #, if available:*

*Description of changes:*


aqjune-aws:
It seems the CI checks hit the memory limit. Reducing the # cores from 50 to 45...

jargh (approved):

PR #159, opened 2024-11-12
Use native OCaml compiler to build & run proofs

aqjune-aws:
This patch compiles proofs using native OCaml compiler, and updates `make proofs` to run the newly created `*.native` executable files.

The new `tools/build-proof.sh` script builds a specific `.ml` proof file. It inlines all loads/needs in the `.ml` file using HOL Light's `inline_load.ml`.
Then, it collects all specifications in the original `.ml` file using already existing `tools/collect-specs.sh`, and adds a print statement at the end of the file, to make its standard output pass the proof checker script. It also adds timer and prints the running time of the proof. Then it is compiled with `ocamlopt.byte`.

Building all proofs can be done with `make build_proofs`. This uses a lot of memory, so compared to the previous version `-j` number given to `make` must be reduced.

This also makes the prerequisite lists tidier using the '*.native' items. Each '.native' file is considered as depending on the '.o' file. This makes compilation of a `.native` file having a long dependency chain take a slightly long time (e.g., p256_montjscalarmul), but I thought this was acceptable.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.a
 *.correct
+*.native
 *.o
 *.obj
 .vscode
--- a/arm/Makefile
+++ b/arm/Makefile
@@ -389,7 +389,7 @@ OBJ = $(POINT_OBJ) $(BIGNUM_OBJ)
 
 libs2nbignum.a: $(OBJ) ; ar -rc libs2nbignum.a $(OBJ)
 
-clean:; rm -f libs2nbignum.a */*.o */*/*.o */*.correct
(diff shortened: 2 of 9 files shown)


aqjune-aws:
Arm proof check is now taking 3hr 55mins (previously 7hr 20mins).
x86 proof check is 2hr 45mins (previously 5hr 10mins). The speedups look fine. :)

jargh (approved):
This is great, thank you -- we will now get feedback from proof build attempts much faster!

PR #160, opened 2024-11-19
Avoid underflow of negative constant immediates in load/stores

aqjune-aws:
This patch prevents negative constant immediates in loads and stores like `ldp _, _, [_, -16]` from introducing a constant that is a result of underflow, e.g., `18446744073709551600`. Instead, `-16` is represented as `word_sub _ (word 16)`.

In addition, this canonicalizes `word_sub (word_add baseptr ofs) ofs2)` to `word_add baseptr (word_sub ofs ofs2)` whenever possible because this makes symbolic analysis on memory instructions with negative constants easier because the base pointers appear clearly after this canonicalization.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/arm/proofs/arm.ml
+++ b/arm/proofs/arm.ml
@@ -272,11 +272,8 @@ let ARM_EXEC_CONV =
                    bytes128 (word_add b (word 0)) = bytes128 b`,
                   REWRITE_TAC[WORD_ADD_0])
   and rth = prove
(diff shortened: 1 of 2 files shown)


jargh (approved):
Thank you, this clearly improves some of the clumsy handling of negative offsets. I will merge it in now.

PR #162, opened 2024-11-22
Update NONOVERLAPPING_TAC to successfully process subtractions

aqjune-aws:
This patch updates `NONOVERLAPPING_TAC` to successfully deal with subtractions. An example that can be now resolved with it is as follows:

```
g `nonoverlapping (x:int64,k) (y:int64,k2) /\ 4 <= i /\ i < k
   ==> nonoverlapping (word_add x (word_sub (word i) (word 4)), 4) (y,k2)`;;

e(REWRITE_TAC[NONOVERLAPPING_CLAUSES]);;
e(STRIP_TAC);;
e(NONOVERLAPPING_TAC);;
```

To achieve this, this patch (1) updates the normalizer to convert `word_sub (word x) (word y)` to `x - y` if `y <= x`, and (2) makes `cmp` eventually rely on `prove_le` rather than quickly raising a failure. Invoking `prove_le` can be quite expensive though; Therefore this patch also has an implication which is making `NONOVERLAPPING_TAC` a step toward being even more 'automatic' while being more expensive.

Orthogonally, this patch updates the assembly simplifier tactic to recognize assumptions of the form `.. |- x = x` and filter them out because this can cause infinite loops in its following ASM_REWRITE tactics.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/common/components.ml
+++ b/common/components.ml
@@ -2879,24 +2879,41 @@ let (NONOVERLAPPING_TAC:tactic) =
     and pth_word_add = (UNDISCH_ALL o prove)
      (`a = word a' ==> b = word b' ==> (word_add a b:int64) = word (a' + b')`,
       REPEAT (DISCH_THEN SUBST1_TAC) THEN REWRITE_TAC [WORD_ADD])
+    and pth_word_sub = (UNDISCH_ALL o prove)
+      (`a = word a' ==> b = word b' ==> b' <= a' ==>
+      (word_sub a b:int64) = word (a' - b')`, MESON_TAC [WORD_SUB])
     and pth_add = (UNDISCH_ALL o prove)
      (`(a == a') (mod (2 EXP 64)) ==> (a + b == a' + b) (mod (2 EXP 64))`,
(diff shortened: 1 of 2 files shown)


aqjune-aws:
I will compare the running times of proofs after CI checks are done, to confirm whether there is a noticable slowdown.

aqjune-aws:
[pr162-arm.xlsx](https://github.com/user-attachments/files/17878481/pr162-arm.xlsx)
[pr162-x86.xlsx](https://github.com/user-attachments/files/17878482/pr162-x86.xlsx)
The result is somewhat a mixture of speed ups and slow downs; some proofs became faster a lot whereas some proofs went slower. The total sum of running times looks fine; there was no significant and consistent slowdown.

jargh (approved):
Thanks, it's very nice to fix the flawed handling of "word_sub". I was initially a bit worried about the efficiency impact, but given the observed results that seems not to be a significant concern :-)

aqjune-aws:
Thanks! :)

PR #163, opened 2024-11-22
Use Compute to evaluate Arm's decode function

aqjune-aws:
This reimplements the `PURE_DECODE_CONV` with the Compute module of HOL Light.
To use the latest `*_COMPUTE_CONV` functions, this bumps the version of HOL Light to the latest one. :)

The main part of this patch was prototyped by @monadius.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/arm/proofs/decode.ml
+++ b/arm/proofs/decode.ml
@@ -896,367 +896,41 @@ let ALIAS_CONV =
   OPERAND_ALIAS_CONV THENC ALIAS_CONV;;
 
 let PURE_DECODE_CONV =
-  let constructors =
-    let constructors_of =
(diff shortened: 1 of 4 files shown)


aqjune-aws:
There is around 2% slowdown in the Arm proofs, but I think this is acceptable. :)

jargh (approved):
Indeed, this is dramatically simpler, removing a source of additional complexity when extending the decoder. In the light of how much better it is from a software engineering standpoint, a small performance overhead is perfectly acceptable, since decoding is not performance-critical anyway. Many thanks to you and of course to Alexey!

Issue #161, opened 2024-11-22
Native compilation of cosimulators


aqjune-aws:
The cosimulators (random tester of the formal semantics) must be natively compiled; there was a patch that introduced the native OCaml compiler this for running proofs, but cosimulators were not patched.

PR #168, opened 2024-12-02
Compile simulators for speed

aqjune-aws:
This patch natively compiles simulators for speed and runs them.

Resolves https://github.com/awslabs/s2n-bignum/issues/161 .

*Issue #, if available:*

*Description of changes:*


jargh (approved):
This is all great, thank you! I have checked that the simulators are being run natively and as a result get through more instructions in the allocated time. I've also tried deliberately breaking the semantics of a couple of instructions and rerunning, and it was correctly flagged with an overall failure of the "make sematest" target.

PR #171, opened 2024-12-06
Adding Cryptographic AES intrinsics for Armv8

pennyannn:
*Description of changes:*

This PR adds modeling of Cryptographic AES intrinsics in preparation for AES-XTS verification. This includes AESE, AESMC, AESD, AESIMC 


arm/proofs/aes.ml
```diff
@@ -0,0 +1,393 @@
...
+
+needs "Library/words.ml";;
+needs "compute.ml";;
+
+(** TODO:define a recursive definition for word_join_list and its CONV *)
+
+let pp_print_num fmt tm =
+  let n = dest_numeral tm in
+  pp_print_string fmt (string_of_num_hex n) in
+install_user_printer("pp_print_num",pp_print_num);;
```
aqjune-aws, line 14: Are all numerals printed as hex after this point, even for proofs not using aes?

arm/proofs/aes.ml
```diff
@@ -0,0 +1,393 @@
...
+time (REWRITE_CONV [input] THENC aese_CONV) `aese (word 0xae6910a45715645a02502baaf5a826c9) (word 0xec882f3270973907d69635eea82d71)`;;
+time (REWRITE_CONV [input] THENC aesmc_CONV) `aesmc (word 0xae6910a45715645a02502baaf5a826c9)`;;
+
+
+
+(* let aes_inv_sub_bytes = new_definition ``;;
+
+let aes_inv_shift_rows = new_definition ``;;
+
+let aes_inv_mix_columns = new_definition ``;; *)
```
aqjune-aws, line 381: What do you think about removing comments here? :)

arm/proofs/aes.ml
```diff
@@ -0,0 +1,393 @@
...
+  ; word 0xa89f3c507f02f94585334d43fbaaefd0
+  ; word 0xcf584c4a39becb6a5bb1fc20ed00d153
+  ; word 0x842fe329b3d63b52a05a6e1b1a2c8309
+  ; word 0x75b227ebe28012079a059618c323c704
+  ; word 0x1531d871f1e5a534ccf73f362693fdb7
+  ; word 0xc072a49cafa2d4adf04759fa7dc982ca
+  ; word 0x76abd7fe2b670130c56f6bf27b777c63
+  ]`;;
+
+TEST [GF2] `GF2`;;
```
aqjune-aws, line 70: What about using `assert` to describe what its expected result would be?

aqjune-aws: Same for other `TEST` uses below too. :)

arm/proofs/aes.ml
```diff
@@ -0,0 +1,393 @@
...
+
+(* ========================================================================= *)
+(* AESMC                                                                     *)
+(* ========================================================================= *)
+let aesmc = new_definition `aesmc (n: 128 word) = aes_mix_columns n`;;
+
+
+(************************************************)
+(** CONVERSIONS FOR AESE and AESMC             **)
+(************************************************)
```
aqjune-aws, line 286: I usually leave thanks to the writer who wrote a prototype of a code, but... probably this is my style of communication. :)

arm/proofs/aes.ml
aqjune-aws: Or instead of using `assert` you can also manually check the output and raise an exception with `failwith "some concrete error message"` if this is easier to debug. :)

arm/proofs/base.ml
```diff
@@ -54,6 +54,14 @@ extra_word_CONV :=
          SHA512SU1_REDUCE_CONV]
         @ (!extra_word_CONV);;
 
+(* ------------------------------------------------------------------------- *)
+(* Additional Cryptographic AES intrinsics                                   *)
+(* ------------------------------------------------------------------------- *)
+
+loadt "arm/proofs/aes.ml";;
+
+extra_word_CONV := [aese_CONV; aesmc_CONV] @ (!extra_word_CONV);;
```
aqjune-aws, line 63: For `aese_CONV` and `aesmc_CONV`, should we follow the style of SHA256/SHA512 which starts with capital letters and explicitly describes that they are REDUCE?

arm/proofs/aes.ml
pennyannn: Sorry, this PR is not ready yet. I'll turn it from draft to official once it's done.

aqjune-aws: oops, sorry. okay!

pennyannn: NP. Sorry for the confusion.

pennyannn: Moved to a new file. Now it will only be used when installed.

arm/proofs/aes.ml
pennyannn: Finished1

arm/proofs/aes.ml
pennyannn: This function only does evaluation of a HOL-Light term and does not check result. I changed its name to `EVAL`, which I think is more appropriate and moved it to a new file called `utils.ml`. I'm using it for debugging purposes during development, but not for adding assertions. I also commented the tests out because some of them are a bit slow and I don't want to make CI run longer. Let me know if you want me to remove those comments.
Also, I like the idea of having an ASSERT function. I could add one to the new file `utils.ml` if requested.

arm/proofs/aes.ml
pennyannn: Added!

arm/proofs/base.ml
pennyannn: What is the difference between conversions that have REDUCE and ones that don't have?

arm/proofs/aes.ml
pennyannn: I've further thought about it. I think I will remove these tests, and I will keep utils.ml for future development use. The reason I want to remove them is because running them would slow down the CI and they are already sort of tested in random instruction simulation. Leaving them in comments is a bad idea because it won't evolve with the code changes.

arm/proofs/base.ml
pennyannn: Done.

jargh (approved):
This all looks great to me now, thank you! I have cross-checked the decoding and some parts of the AES definition and have run the simulator a bit more focusing on just the new instructions. Also, the final versions of the conversions are pretty efficient, and they fail very quickly when inapplicable thanks to the top-level pattern match. I'll approve and merge :-)

PR #172, opened 2024-12-07
Avoid access to stack pointer + negative offset in p384_montjadd

aqjune-aws:
This is a patch that fixes p384_montjadd and p384_montjscalarmul (which has p384_montjadd inlined) to avoid memory access to stack pointer + negative offset in the middle of the function.
This is because when a process is interrupted, in principle, context switching to a handler may overwrite data under the stack pointer: https://forum.osdev.org/viewtopic.php?t=21720
This also caused valgrind from AWS-LC side to fail.

This is porting the patch of @dkostic:
https://github.com/aws/aws-lc/pull/2036/commits/65808f80b26171e2d66347b84f99c1acb26efee6 .. and also updates the proof according to it.
Also, this adds valgrind to the test so that it can print such cases.

*Issue #, if available:*

*Description of changes:*


tests/test.c
```diff
@@ -7671,6 +7671,7 @@ int test_bignum_mux16(void)
      i = rand() & 15;
      reference_copy(k,b1,k,bs+k*i);
      bignum_mux16(k,b2,bs,i);
+     free(bs);
```
aqjune-aws, line 7674: This is a patch received from John. :)

aqjune-aws:
arm/proofs/p384_montjscalarmul.ml has a large diff because `save_literal_from_elf` prints immediate offsets in `word`, not `iword`. It did not touch any other part.

jargh (approved):
Thank you very much for the quick fix :-)

PR #173, opened 2024-12-10
Implement IBT on amd64 / i386

lgv5:
*Description of changes:* as the title reads, this sprinkles `endbr64` / `endbr32` as needed in the assembler files to be able to run with CET enabled. I tested this on an i7 1165G7 on OpenBSD, as CET is enabled by default there. I do get SIGILLs without the patch, and the testsuite passes with it. The PR affects quite a lot of files as it basically adds a `_CET_ENDBR` after each function definition. The following script was used for patching the files:

```
#!/bin/sh
find . -type f -name '*.S' -exec grep -lx 'S2N_BN_SYMBOL(.*):' {} + |
    while read -r f; do
        sed -i -f /dev/stdin "$f" <<'EOF'
/^S2N_BN_SYMBOL([^)]*):$/ {
        :Loop
        n
        /^S2N_BN_SYMBOL([^)]*):$/ b Loop
        i\
        _CET_ENDBR
}
EOF
done
```

The `sed` machinery is only needed to avoid adding multiple `_CET_ENDBR`s in definitions like

```
S2N_BN_SYMBOL(curve25519_x25519base):
S2N_BN_SYMBOL(curve25519_x25519base_byte):
```


lgv5:
Alternatively, I can introduce

```
#define S2N_BN_SYM_FUNCTION(name) \
S2N_BN_SYMBOL(name): \
  _CET_ENDBR
```

in `include/_internal_s2n_bignum.h` and replace all the stray `S2N_BN_SYMBOL(name):` with it, but that also doesn't escape of the complexity of a 629-files diff. The upside of this, tho, is that future additions wouldn't have to worry about `_CET_ENDBR` as long as they use the new define.

jargh:
Thank you for addressing this whole issue, which as far as I know had passed us by completely. We should probably be doing more testing on BSD versions!

Your solution seems to be a nice lightweight one in the sense that the additional ENDBR instructions are only added with `__CET__` enabled, and otherwise things are unchanged. This has the pleasant effect that the proofs don't need to be changed :-)

Of course, it is unfortunate that the proofs in principle don't cover the CET-enabled variants. We could adopt a more heavyweight solution of just adding ENDBR instructions regardless, trusting them to have minimal performance impact and treating them as NOPs. According to the Intel doc quoted in the discussion here they can safely be modeled as a NOP and assumed to behave as such even on legacy processors:

https://stackoverflow.com/questions/56905811/what-does-the-endbr64-instruction-actually-do

My inclination is to go with your solution first and maybe think about the more far-reaching change in the future. Any other opinions?

lgv5:
> Thank you for addressing this whole issue, which as far as I know had passed us by completely. We should probably be doing more testing on BSD versions!

It was a pleasure, compared to other projects where I did IBT patching. The only bothersome part is the amount of files to patch. :D

AFAIK, only OpenBSD enforces CET, with Intel CPUs from gen 11th and onwards. Also, we only do IBT, not full CET (we don't do shadow stacks).

> My inclination is to go with your solution first and maybe think about the more far-reaching change in the future. Any other opinions?

After I thought more about it, from a maintainer point of view, I think the approach of the comment (defining a macro that outputs a function preamble) is better, in the sense that there is no need for another 629-files patch. But well, I'm not a maintainer. I'm still up for biting the bullet again, if you prefer that approach.

lgv5:
Hello @jargh ! Are there any updates in here? We really we can merge it so it can eventually backpropagate to aws-lc-rs/aws-lc-sys, as more things in Rust land are depending on it.

jargh:
Sorry for dropping the ball on this. I don't think we'll have time to go for the more complex solution of actually formally modeling the `endbr` instructions very soon (i.e. for a few months). So in the light of that, I think we'll go, at least for the short term, with your solution pretty much "as is", i.e. modifying the syntax when the appropriate build flags are set. Would you mind checking if it needs rebasing w.r.r. any other updates and then I'll prioritize merging it in.

lgv5:
Hey jargh! I rebased the PR onto `main`, and I added a reentrant version of the script, which should help to avoid missing `_CET_ENDBR` instructions. It can become specially handy on CI: I believe that adding

```
sh add-ibt-endbr.sh && git diff --exit-code .
```

after https://github.com/awslabs/s2n-bignum/blob/main/codebuild/tests.yml#L12 should do the trick. Feel free to drop it if it isn't desired.

jargh:
Many thanks! I'm working on assimilating and checking it. I'll also experiment a bit to see what the performance hit seems to be, because if it's not much (and assuming there are no other issues) we really may want to prioritize making this the default and updating the proofs accordingly.

jargh:
Since all our code is assuming 64-bit mode, I suspect we're safe ignoring `ENDBR32`, right? (For example, not including it in `x86/allowed_asm`.) Or would it still be relevant in some situation?

I'm actually somewhat warming to the idea of making this the default and updating the proofs, but let me do a bit more investigating and poll some other people about what they think.

jargh:
I've set up a branch adopting your changes and also adding a NOP-assuming semantics to the formal model.
https://github.com/jargh/s2n-bignum-dev/tree/cti
So far so good. Next I'll see if the fixup of the proofs if I enable IBT by default is manageable. I foresee some messiness, but maybe of a systematic enough character that it won't be too bad.

jargh:
OK, I have finally elaborated this PR into one that also modifies the proofs: https://github.com/awslabs/s2n-bignum/pull/203 - I'd be interested to know what you think. The main points are:

* `ENDBR64` is now added by default regardless of platform choices, and it is stated as an explicit byte sequence to make sure the code at least builds and runs using older compilers/assemblers that don't recognize it.
* The proofs are updated so that each result now has 4 variants for the product of the choices {IBT,no_IBT} x {Standard_ABI,Windows_ABI}, and it is the two IBT ones (i.e. with `ENDBR64` at the beginning) that are explicitly checked against the generated object code.

lgv5:
I'm not sure about replacing the inclusion of cet.h with only the definition of `_CET_ENDBR`. Below is the cet.h on my machine. It does something with sections, which I don't really know whether it's important or not. Maybe you can use the following machinery instead, which should still output the bytes even when the toolchain doesn't support it, and add section when the toolchain is aware of CET.

```
#if defined(NO_IBT)
#define _CET_ENDBR
#elif defined(__CET__)
#include <cet.h>
#else
#define _CET_ENDBR .byte 0xf3,0x0f,0x1e,0xfa
#endif
```

`cet.h`:

```
/*===------ cet.h -Control-flow Enforcement Technology  feature ------------===
 * Add x86 feature with IBT and/or SHSTK bits to ELF program property if they
 * are enabled. Otherwise, contents in this header file are unused. This file
 * is mainly design for assembly source code which want to enable CET.
 *
 * Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 * See https://llvm.org/LICENSE.txt for license information.
 * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 *
 *===-----------------------------------------------------------------------===
 */
#ifndef __CET_H
#define __CET_H

#ifdef __ASSEMBLER__

#ifndef __CET__
# define _CET_ENDBR
#endif

#ifdef __CET__

# ifdef __LP64__
#  if __CET__ & 0x1
#    define _CET_ENDBR endbr64
#  else
#    define _CET_ENDBR
#  endif
# else
#  if __CET__ & 0x1
#    define _CET_ENDBR endbr32
#  else
#    define _CET_ENDBR
#  endif
# endif


#  ifdef __LP64__
#   define __PROPERTY_ALIGN 3
#  else
#   define __PROPERTY_ALIGN 2
#  endif

	.pushsection ".note.gnu.property", "a"
	.p2align __PROPERTY_ALIGN
	.long 1f - 0f		/* name length.  */
	.long 4f - 1f		/* data length.  */
	/* NT_GNU_PROPERTY_TYPE_0.   */
	.long 5			/* note type.  */
0:
	.asciz "GNU"		/* vendor name.  */
1:
	.p2align __PROPERTY_ALIGN
	/* GNU_PROPERTY_X86_FEATURE_1_AND.  */
	.long 0xc0000002	/* pr_type.  */
	.long 3f - 2f		/* pr_datasz.  */
2:
	/* GNU_PROPERTY_X86_FEATURE_1_XXX.  */
	.long __CET__
3:
	.p2align __PROPERTY_ALIGN
4:
	.popsection
#endif
#endif
#endif
```

jargh:
Yes, good point, there's more to CET than just the endbr64 instructions. I'll switch to the formulation you suggest, using the header if __CET__ is defined and falling back on the manual definition otherwise.

jargh:
See https://github.com/awslabs/s2n-bignum/pull/203/commits/c49d9b1b50415d11c78a3f7a79411e898e07016f which applies this tweak.

lgv5:
I managed to try the hardcoded bytes version and it still ran fine in OpenBSD (which should trigger a SIGILL without the landing pad, as all the functions are called by pointers in the tests). The IBT bits look good to me.

I can't provide any input on the proofs.

- - -

Should we close this PR and follow-up in the other, if needed?

jargh:
It's probably safe to close this one, yes. I'll open the other one up to review by the various local AWS-LC people and if they are happy we can merge it in. Thanks for your patience on this...

lgv5:
Work continues in #203 . :wave:

PR #175, opened 2024-12-11
Make symbolic simulation for equiv checking infer output of loads

aqjune-aws:
This patch updates symbolic simulation tactics for equivalence checking to infer the output of memory loads from assumptions if there is no directly matching equality.

For example, if there is an assumption `read (memory :> bytes (ptr,8*n)) = a` but nothing about `read (memory :> bytes64 (word_add (ptr,8*k)))`, it tries to construct its output as `bigdigit a k` using `SIMPLE_ARITH_TAC`. This helps remove boilerplate codes in proofs that manually construct all equalities between 64-bit and 128-bit memory reads at known locations, which typically looked like:
```
REPEAT (FIRST_X_ASSUM BIGNUM_EXPAND_AND_DIGITIZE_TAC) THEN
ASM_PROPAGATE_DIGIT_EQS_FROM_EXPANDED_BIGNUM_TAC THEN
(* necessary to run ldr qs *)
COMBINE_READ_BYTES64_PAIRS_TAC THEN
```

This also reduces the sizes of assumptions, which may affect positive results to proof checking times, but in fact caused slowdown to larger equivalence checking tactics such as p256/p384's point operations, hence they are manually written.

The main tactics are `MK_MEMORY_READ_EQ_BIGDIGIT_CONV` and `DIGITIZE_MEMORY_READS_TAC`. They construct equations involving the unseen memory reads and `bigdigit`. They are passed to `ARM_N_STEP_TAC`, which are called 'auxiliary memory read equations, and they are either `ASSUME_TAC`ed or passed to the caller separately from other normal equations describing the actual outputs of a simulated instruction.

On top of these updates, I also added a few utility functions like `get_input_output_regs` which gets the input and output register from a basic block. These will be used in my future patches.

*Issue #, if available:*

*Description of changes:*

*Issue #, if available:*

*Description of changes:*


jargh (approved):
This all looks good, many thanks! And we may well want to make similar changes to symbolic simulation generally in the near future if the performance impact is not a problem, since it will save a lot of the inconvenience caused by manually modifying assumptions about memory.

PR #177, opened 2024-12-20
Prefix all local labels with file name

dkostic:
*Issue #, if available:*
N/A

*Description of changes:*
Prefix all local labels with the file name they are defined in.
Done with this small script:
```
#!/bin/bash

find arm x86 x86_att -type f -name "*.S" | grep -v -E "proof|tutorial|unopt" > files.txt

while read -r target_file; do
    grep -o '^[^/^)]*:' ${target_file} | sed 's/://' > labels.txt
    prefix=$(echo ${target_file} | awk -F'[/.]' '{print $(NF-1)}')
    while read -r label; do
        if [[ "${label}" == "${prefix}"* ]]; then
            continue
        fi
        sed -i "/^\/\//!s/\b${label}\b/${prefix}_${label}/g" ${target_file}
    done < labels.txt
done < files.txt

rm files.txt labels.txt
```
I couldn't use the existing script (https://github.com/awslabs/s2n-bignum/blob/main/tools/per_file_unique_labels.py) because it prefixes already prefixed labels (I was too lazy to figure out how it works and fix it so I wrote my small bash script above).


jargh:
Thank you, it'll make future FIPS builds so much easier if we just do this definitively once and for all! One little nitpick is that it's occasionally picking up words in comments, so it might be worth taking an extra pass over those instances, e.g.
`// ready for the next time round the bignum_cdiv_exact_loop.
`
Otherwise we would expect/hope no actual changes in the code besides the label names and hence no changes to the proofs - the CI check will confirm that.

dkostic:
> Thank you, it'll make future FIPS builds so much easier if we just do this definitively once and for all! One little nitpick is that it's occasionally picking up words in comments, so it might be worth taking an extra pass over those instances, e.g. `// ready for the next time round the bignum_cdiv_exact_loop. ` Otherwise we would expect/hope no actual changes in the code besides the label names and hence no changes to the proofs - the CI check will confirm that.

Done! I updated the script to do
```
sed -i "/^\/\//!s/\b${label}\b/${prefix}_${label}/g" ${target_file}
```
 instead of
```
sed -i "s/\b${label}\b/${prefix}_${label}/g" ${target_file}
```

jargh (approved):
All now looks very good, thank you! I made a few checks and have in particular confirmed that all labels throughout the respective arm and x86 codebases are now distinct.

PR #178, opened 2024-12-26
Add theorems and tactics for ensures_n.

bdalrhm:
This PR splits `common/relational2.ml` into two: `common/relational_n.ml` and `common/relational2.ml`. It additionally adds basic theorems and tactics about `eventually_n` and `ensures_n` similar to those for `eventually` and `ensures`.


aqjune-aws:
I talked with Abdal and this change will be included in my upcoming pull request with his credit mentioned, because the equivalence checking proof will use these lemmas and tactics.

PR #179, opened 2024-12-28
Add `bignum_emontredc_8n_cdiff`

aqjune-aws:
This patch adds a verified implementation of Montgomery reduction, `bignum_emontredc_8n_cdiff`. This is an optimized version of `bignum_emontredc_8n` for Arm with slow multipliers such as Graviton2. This incorporates three optimizations:
(1) Expensive multiplications are properly vectorized with NEON instructions, as done in NIST curves field operations (2) The inner loop is software pipelined, and instructions in basic blocks are rescheduled, both of which are done using SLOTHY (3) Multiplied differences between adjacent words of modulus, which are used in ADK, are cached in a temporary buffer and reused in the inner loop.

The benchmark result on Graviton2 is as follows:

```
bignum_emontredc_8n (64 -> 32)  :  2027.5 ns each (var  0.0%, corr -0.10) =     493215 ops/sec
bignum_emontredc_8n_neon (64 -> 32):  1605.1 ns each (var  0.1%, corr -0.08) =     623014 ops/sec
bignum_emontredc_8n_cdiff (64 -> 32):  1294.7 ns each (var  0.1%, corr -0.03) =     772365 ops/sec <-- this
```

The intermediate program that is a variant of `bignum_emontredc_8n` with optimizations (1) and (3) applied but not (2) is stored at `arm/fastmul/unopt/bignum_emontredc_8n_cdiff_base.S`. Both `bignum_emontredc_8n_cdiff.S` and `bignum_emontredc_8n_cdiff_base.S` are written by Hanno Becker. This is a patch that delivers his implementation, with the formal correctness proof of `bignum_emontredc_8n_cdiff.S` attached.

#### Reusing cached difference across different Montgomery reductions

In theory, the cached difference can be even reused across different Montgomery reduction calculations if the same modulus is used. This can make the amortized speed of Montgomery reduction even faster. This patch does not provide functions for this type of use case yet. The difference calculation cacheing is embedded in the new function and unconditionally run.

#### Additional constraints on parameter k

`bignum_emontredc_8n_cdiff` has two more constraints on k than `bignum_emontredc_8n`. First, k < 2^32. This bounds the maximum size of the buffers z/m/m_precalc, always leaving some amount of empty space in the memory since memory size is 2^64. This is necessary when proving program equivalence, e.g., `p ~= p'`, by composing two other equivalences `exists p''. p ~= p'' /\ p'' ~= p'`. It needs the existence of `p''` which will be placed in the memory, and this requires an empty memory space for that. Second, 16 <= k. This is because applying software pipelining unrolls the inner loop of Montgomery reduction. One possible solution for removing this constraint is to check whether k < 16 and then simply call the old `bignum_emontredc_8n`. I chose to leave this as a precondition because I thought that avoiding redundancies between implementations sounded like a good idea... (and this will add its redundancy).

#### Other updates

- As discussed in https://github.com/awslabs/s2n-bignum/pull/178 , common/relational_n.ml is added and the definition of `eventually_n`/`ensures_n` are moved to this file. Also, a few fundamental lemmas and tactics of Abdal are located.
- bignum_emontredc_8n is updated to remove a redundant conditional branch before inner loop.
- `ldr`/`str` with a Q register is added to the decoder: https://developer.arm.com/documentation/ddi0596/2020-12/SIMD-FP-Instructions/LDR--immediate--SIMD-FP---Load-SIMD-FP-Register--immediate-offset--
- `DECODE_CONV` is updated to not use `word_compute_add_convs` because it was too aggressively reducing expressions, esp., `iword (const)`.
- Minor edits in several files are from renaming `prove_correct_n` to `prove_ensures_n`.

*Issue #, if available:*

*Description of changes:*


jargh (approved):
This all looks great! The proof is a very complex one, but all that effort was well worth it based on the performance results, which I have confirmed. Many thanks!

PR #180, opened 2025-01-02
Add instructions LD1/ST1 and LD2/ST2

pennyannn:
*Description of changes:*

This PR adds Arm instructions LD1/ST1  (1 register, post-immediate offset) and LD2/ST2 (2 registers, post-immediate offset). 
In addition, a unit proof file is added for proving correctness properties of these two sets of instructions. This helps relieve the lack of conformance testing for these load store instructions.


arm/proofs/instruction.ml
```diff
@@ -1931,6 +1948,193 @@ let arm_STP = define
...
+    \s. let address = read Rn s in
+        let eaddr = word_add address (offset_address off s) in
+        (if (Rn = SP ==> aligned 16 address) /\
+            (offset_writesback off ==> orthogonal_components Rt Rn)
+         then
+           Rt := read (memory :> wbytes eaddr) s ,,
+           (if offset_writesback off
+            then Rn := word_add address (offset_writeback off)
+            else (=))
+         else ASSIGNS entirety) s`;;
```
aqjune, line 1963: This can be a naive question, but if arm_LD1_1 is simply loading the whole word can using existing `arm_ldst*` be used? This will make the diff of this patch smaller & have less maintenance cost.

arm/proofs/ldst.ml
```diff
@@ -0,0 +1,159 @@
+(*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT-0
+ *)
```
aqjune, line 4: `ldst.ml` seems to contain sanity 'proof' tests for the new instructions, which is similar to what simulator does in the sense that it inspects the concrete output of the instructions. What about moving these proofs to simulator.ml, or even create a new simulate_ldst.ml & move these proofs to there & let simulator.ml load simulate_ldst.ml?
This can also help avoiding running these proofs everytime whenever `base.ml` is loaded; base.ml is supposed to contain lemmas and tactics that can be reused for general purposes, but it seems proofs in ldst.ml cannot be used that way.

pennyannn: Thanks for the recommendation! I was having doubt myself if base.ml is the right place for it. I will update the PR.

arm/proofs/instruction.ml
pennyannn: In general, LD1/ST1 is not the same as arm_LDR/STR because they are also defined for 2,3,4 registers. However, I don't see any use of those cases in my program, so I only modeled the case for 1 register. For 1 register, it is exactly the same as LDR and STR. I merged the definitions in the following commit. Let me know if you like it :)

aqjune: The two updates looked really nice, thanks..! :)

aqjune:
Not a critical problem, but I wonder whether the commit history can be squashed into one commit so that the intermediate failures can disappear. Having the intermediate commits removed is really useful when I want to bisect what is the latest commit that makes some test/proof successful.

pennyannn:
> Not a critical problem, but I wonder whether the commit history can be squashed into one commit so that the intermediate failures can disappear. Having the intermediate commits removed is really useful when I want to bisect what is the latest commit that makes some test/proof successful.

Good idea! Done squashing.

aqjune:
I am still seeing multiple commits in this PR - probably a few latest commits are squashed but not all.

jargh (approved):
All looks great now, and thanks for incorporating a few offline suggestions

PR #182, opened 2025-01-14
Add LD1R, FMOV and SRI instructions

pennyannn:
*Description of changes:*

Adding a couple more Arm instructions used in SHA3 and AES-XTS:
1. LD1R post-immediate offset 
2. FMOV (general) double-precision
3. SRI (vector)

### DIFF
--- a/arm/proofs/decode.ml
+++ b/arm/proofs/decode.ml
@@ -347,7 +347,7 @@ let decode = new_definition `!w:int32. decode w =
     SOME (arm_ldst2 is_ld Rt (XREG_SP Rn) (Postimmediate_Offset (word 32)) 128 esize)
 
   // LD1R, Post-immediate offset, size 64 and 128
-  | [0b0:1; q; 0b001101110:9; Rm:5; 0b1100:4; size:2; Rn:5; Rt:5] ->
+  | [0b0:1; q; 0b001101110111111100:18; size:2; Rn:5; Rt:5] ->
     let esize = 8 * (2 EXP (val size)) in
     let datasize = if q then 128 else 64 in
     let off = word (esize DIV 8) in
--- a/arm/proofs/decode.ml
+++ b/arm/proofs/decode.ml
@@ -426,11 +426,7 @@ let decode = new_definition `!w:int32. decode w =
       else
         let immb = abc in
         let Rn = defgh in
-        let highest_set_bit =
-          if bit 3 immh then 3 else
-          if bit 2 immh then 2 else
-          if bit 1 immh then 1 else 0 in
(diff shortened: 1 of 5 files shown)


jargh:
I think it may be possible to use the relatively new `word_duplicate` function in the HOL Light libraries to avoid the multiple analogous functions `word_replicate2`, `word_replicate4` etc.

pennyannn:
> I think it may be possible to use the relatively new `word_duplicate` function in the HOL Light libraries to avoid the multiple analogous functions `word_replicate2`, `word_replicate4` etc.

Thanks for mentioning it! Updated.

arm/proofs/decode.ml
```diff
@@ -313,6 +313,13 @@ let decode = new_definition `!w:int32. decode w =
   | [0b00:2; 0b1111001:7; is_ld; 0:1; imm9:9; 0b00:2; Rn:5; Rt:5] ->
     SOME (arm_ldst_q is_ld Rt (XREG_SP Rn) (Immediate_Offset (word_sx imm9)))
 
+  // LD1R, Post-immediate offset, size 64 and 128
+  | [0b0:1; q; 0b001101110:9; Rm:5; 0b1100:4; size:2; Rn:5; Rt:5] ->
```
pennyannn, line 317: Reminder from John: LD1R only handles immediate offset, constrain `Rm` to be `0b11111`.

jargh (commented):
This all looks absolutely functionally correct, and passes pretty intense testing -- thanks for adding the full cosimulation including LD1R! I have a couple of stylistic suggestions that you might want to consider before we merge:

* It would be more consistent with the general style to change the argument order of `arm_FMOV_FtoI` and `arm_FMOV_ItoF` such that `Rd` is the first argument, since pretty much all the other `arm_XXX` follow this approach and it makes the "abstract syntax" you get from decoding a bit more like the usual ARM dest := sources style.
* There could also be a case for using the `word_insert` library function in the more complex `arm_FMOV_ItoF` case of inserting into `D[1]`. It might only look marginally shorter but I think it will work better with other simplification functions like `WORD_SIMPLE_SUBWORD_CONV` when used in proofs.

jargh:
Oh, actually one other minor stylistic comment that applies to both the new `SRI` and the existing `SLI` and `USRA`, which are so analogous. In place of the if-then-else tower for `highest_set_bit` there is the possibility of using `3 - word_clz immh`. Possibly it looks slightly arcane, but maybe it's actually clearer, and that is also how it's done just above in `USHR` as well as a couple of others further down.

jargh (approved):
All looks great now, thanks! I will merge.

PR #186, opened 2025-01-30
Enable memory operations in ARM cosimulator

jargh:
This enables cosimulation of LD1/ST1/LD2/ST2 in the simulation setup as well as the existing register-to-register instructions. Adding more memory operations should now be relatively routine, although some individual coding is needed to set up self-contained stack-safe test sequences, as done for the present operations in "cosimulate_ldst_12()". The main changes to support memory operations in the simulator are:

- Factor out the main cosimulation code from the random selection of instructions, and also generalize cosimulation to accept any list of instructions, not just a single one.

- Save and check additional state during cosimulation, a 256-byte buffer at SP,...,SP+255. The stack pointer value must be the same before and after, and the results in other registers independent of that value.

- Complicate the tactics a bit in the memory case to re-express memory assignments in a way that makes sure they are applicable, in the final resort reducing everything to byte arrays.

- For memory operations, set up a self-contained test harness that satisfies the constraints above, i.e. maintains original stack pointer and does not write outside [SP],...,[SP+255], but still tests the memory operation over a good range of instances.

There are a few other minor changes that make the code slightly more portable, and in particular make it run cleanly on Mac OS using Clang there is now a "_harness" as well as "harness" entry point name for the test harness, and the printf-ing of uint64_t uses more strictly portable macros.

At the moment, about 10% of runs are the memory operations and 90% are the original register-to-register instructions. The latter should work in much the same way, with only a slight slowdown from the additional infrastructure.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/arm/proofs/simulator.ml
+++ b/arm/proofs/simulator.ml
@@ -181,7 +181,7 @@ and tac_after memop =
 
 (*** Cosimulate a list of ARM instruction codes against hardware.
  *** To pass, the formal simulation has to agree with the hardware,
- *** only modify the 256-byte buffer [RSP,..,RSP+255] and also
+ *** only modify the 256-byte buffer [SP,..,SP+255] and also
  *** leave the final SP value the same as the initial value, though
  *** it can be modified in between.
  ***)
--- a/arm/proofs/simulator.ml
+++ b/arm/proofs/simulator.ml
@@ -28,7 +28,23 @@ let regfile = new_definition
     val(read Q16 s); val(read Q17 s); val(read Q18 s); val(read Q19 s);
     val(read Q20 s); val(read Q21 s); val(read Q22 s); val(read Q23 s);
     val(read Q24 s); val(read Q25 s); val(read Q26 s); val(read Q27 s);
-    val(read Q28 s); val(read Q29 s); val(read Q30 s); val(read Q31 s)]`;;
+    val(read Q28 s); val(read Q29 s); val(read Q30 s); val(read Q31 s);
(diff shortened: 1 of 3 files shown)


arm/proofs/simulator.ml
```diff
@@ -92,54 +109,115 @@ check_insns();;
...
+  (if memop then MAP_EVERY MEMORY_SPLIT_TAC (0--4) else ALL_TAC) THEN
+  ENSURES_FINAL_STATE_TAC THEN ASM_REWRITE_TAC[] THEN
+  (if memop then CONV_TAC(ONCE_DEPTH_CONV READ_MEMORY_MERGE_CONV)
+   else ALL_TAC) THEN
+  ASM_REWRITE_TAC[] THEN extra_simp_tac THEN
+  PRINT_GOAL_TAC THEN NO_TAC;;
+
+(*** Cosimulate a list of ARM instruction codes against hardware.
+ *** To pass, the formal simulation has to agree with the hardware,
+ *** only modify the 256-byte buffer [RSP,..,RSP+255] and also
```
pennyannn, line 184: RSP -> SP

aqjune-aws (approved):
Modulo Yann's comment, the change looks good to me. I could check that the simulator runs LDR,LD2,STR,ST2 successfully.
After fixing the errata, please feel free to merge whenever you are convenient..!

arm/proofs/simulator.ml
jargh: Thanks - I must have been subconsciously anticipating your x86 version :-)

PR #189, opened 2025-02-05
Enable testing of instructions with memory operands in x86_64 simulator

pennyannn:
*Description of changes:*
This is a PR following what is done in https://github.com/awslabs/s2n-bignum/pull/186 for enabling testing of instructions with memory operands in x86_64 simulator.

Specifically, cosimulation tests with memory operands are added for the following instructions:
```
ADD, ADC, OR, SBB, SUB, XOR, MOV, CMOVA, CMOVB, MUL, PUSH and POP
```
They are tested for two addressing modes: `base + scale * index + displacement` and `base + displacement`. The tested operand size is 64 and displacement is `disp8`. One additional test is added when `base` register is `rsp`. They are represented as the following calls in the code:
```
cosimulate_mem_full_harness([opcode]); (* base + scale * index + displacement *)
cosimulate_mem_base_disp_harness([opcode]); (* base + displacement *)
cosimulate_mem_rsp_harness([opcode]); (* rsp_base + scale * index + displacement *)
```
To simplify the problem, the register usage is mostly fixed, however the parameters (`base`, `scale`, `index`, and `displacement`) are randomized. For `PUSH` and `POP`, the register usage is also randomized. 
These instructions and their respective addressing modes are chosen because they appear in existing s2n-bignum assembly. 
The `cosimulate_xxx_harness` functions are added in file x86-mem-insns.ml for convenience of adding tests for future x86 instruction modeling. 

Unlike in Arm, many of the instructions can take memory operands in x86. Therefore I set the cosimulation fraction to be 50% for both non memory-accessing instruction and memory-accessing instruction. The number of tests run in a 40min run is reduced from 2000 to 200, meaning it is now 10 times slower. Reducing the percentage to 10% and around 800 instructions are tested. Some analysis for why this slowdown:
1. The tactics used for the simulation proof is slightly more complex than the ARM version. Specifically, we repeated the use of `READ_MEMORY_MERGE_CONV` and `ASM_REWRITE_TAC[]` as in 
```
(if memop then CONV_TAC(ONCE_DEPTH_CONV READ_MEMORY_MERGE_CONV)
   else ALL_TAC) THEN
  ASM_REWRITE_TAC[] THEN extra_simp_tac THEN
  (if memop then CONV_TAC(ONCE_DEPTH_CONV READ_MEMORY_MERGE_CONV)
   else ALL_TAC) THEN
  ASM_REWRITE_TAC[] THEN extra_simp_tac
```
2. While running conversion over these tests, terms like `val (word xxx))` appear in expressions causing proofs to fail. To fix this problem, John (@jargh) proposes adding a tactic `EPTH_CONV WORD_NUM_RED_CONV THENC` in `x86_conv`. This change has an impact on all tests
3. Auxiliary instructions are slightly more. For example `cosimulate_mem_full_harness` function takes five instructions to simulate one

One minor thing, I often find myself having to test if an instruction could be decoded by s2n-bignum, so I wrote this function that currently resides in simulator.ml:
```
(* A function that decodes a list of bytes into an x86 instruction.
 Could be used for figuring out if an instruction exist in s2n-bignum. *)
let decode_inst ibytes =
  let ibyteterm =
     mk_flist(map (curry mk_comb `word:num->byte` o mk_small_numeral) ibytes) in
  let execth = X86_MK_EXEC_RULE(REFL ibyteterm) in
  let decoded = mk_flist
     (map (rand o rand o snd o strip_forall o concl o Option.get)
       (filter Option.is_some (Array.to_list (snd execth)))) in
  let _ = print_term decoded in
  decoded
```
I find it really helpful in times, but I'm not sure where to put it. Let me know if it should be moved to somewhere else.


aqjune-aws:
Hi Yan, in the x86-sematests log the proof states are being printed when `~undefined` remains in the conclusion. Do you want to remove this logging? This is useful when wants to briefly look at the Codebuild output and see the checkd instructions trace.

pennyannn:
> Hi Yan, in the x86-sematests log the proof states are being printed when `~undefined` remains in the conclusion. Do you want to remove this logging? This is useful when wants to briefly look at the Codebuild output and see the checkd instructions trace.

Thanks, June. The log is removed.

x86/proofs/x86-mem-insns.ml
```diff
@@ -0,0 +1,209 @@
+(*
```
aqjune-aws, line 1: Do you want to update x86/Makefile to add x86-mem-insts.ml as dependency of the 'proofs/simulator.native' target? This will help rebuild 'proofs/simulator.native' when necessary. (In fact, I realized that 'x86-insns.ml' was also missing from its dependency..... :( )

And this is my personal opinion, but the file name of x86-mem-insts.ml doesn't give much intuitions about how the source file is supposed to be used. What about sim-mem-insts.ml that includes 'sim' to hint that this is for simulator? 
Or embed its contents in x86/proofs/simulator.ml instead & remove x86-mem-insts.ml (Makefile update isn't necessary in this case)?

x86/proofs/x86-mem-insns.ml
```diff
@@ -0,0 +1,209 @@
...
+   cosimulate_mem_rsp_harness([0x0F; 0x42]);
+   (* MUL r/m64 *)
+   cosimulate_mul_full_harness();
+   cosimulate_mul_base_disp_harness();
+   cosimulate_mul_rsp_harness();
+   (* PUSH r64 *)
+   cosimulate_push_harness();
+   (* POP r64 *)
+   cosimulate_pop_harness();
+  ];;
```
aqjune-aws, line 209: For non-memory x86 instructions, `list-x86-insns.sh` that collects all such instructions at 'iclasses' list of x86-insns.ml and uses the list to pick which instruction to cosimulate.
Even if random testing will inevitably miss some cases, this still helps us give a kind of claim stating that "if we run the test for a sufficiently long time, we will test all instructions that appeared in bignum functions".

Do you think it is possible to achieve the same thing for x86 memory instructions as well? This is just my curiosity... :)
It is true that we are not doing this for Arm's loads and stores, but this must be expanded in the future as well I believe.

x86/proofs/simulator.ml
```diff
@@ -332,13 +458,29 @@ let run_random_simulation () =
...
+  let icodes = el (Random.int (length mem_iclasses)) mem_iclasses in
+  cosimulate_instructions true icodes;;
+
 (* ------------------------------------------------------------------------- *)
 (* Keep running tests till a failure happens then return it.                 *)
 (* ------------------------------------------------------------------------- *)
 
+let run_random_simulation() =
+  if Random.int 100 < 90 then run_random_regsimulation()
+  else run_random_memopsimulation();;
```
aqjune-aws, line 482: Q: what do you think about counting the number of memory vs. non-memory instructions that are simulated & eventually print at line 497? This can be pretty helpful when we are looking at Codebuild log, to see any changes in cosimulation speed. This did not exist in Arm cosimulator though, but I thought this could be useful (for both).

aqjune-aws (commented):
Hi Yann, thanks for working on this. It was really wonderful to see that this is working on my x86 machine too.
I have a question about two possible solutions for speed improvement:

1. Can we use `{ARM,X86}_VSTEPS_TAC` for instructions only for & after the main memory instruction? This will significantly reduce the size of assumption (hence faster simulation).
2. Would removing assumptions of larger-than-a-byte memory reads `memory :> bytes16 (..)` , `memory :> bytes32 (..)`, ... and only keep `memory :> bytes8 (..)` at the end of
    `tac_before` work? We can achieve this using `DISCARD_ASSUMPTIONS_TAC (fun th -> let t = concl th in ... (* we can use 'term_match' to find a subterm 'memory :> bytes16 x' from t *))`.
    This will also reduce the size of assumptions. In theory the simulator will still work because we are splitting memory reads in `tac_after` as well, but.. who knows :/

If you are busy, maybe I can try writing a patch by myself and add a commit this to PR if you don't mind.
I conjecture these could also have been applied when the ARM version of memory instruction simulation was being added.

x86/proofs/x86-mem-insns.ml
aqjune-aws: To clarify, I don't think this PR needs to address this challenge; it can be a future work of somebody. Maybe what we will need in the close future is to open an github issue that records this fact.

aqjune-aws:
I just pushed two commits that implement my optimizations to x86 simulator. :) According to the Codebuild log, the number of simulated instruction has increased a bit (around 700 to 950?). Hope this can be even faster. Maybe we can even utilize cach in nonoverlapping.ml , IDK. I will push another commit that implements the same optimization on the Arm simulator, finger crossed...

x86/proofs/x86-mem-insns.ml
pennyannn: That is definitely one of the problems I faced when working on this PR. There seems to be two flavors of cosimulation testing in s2n-bignum: one is like what is done in Arm cosimulation where one randomly generate instructions according to some templates; another way is like what is done for non-memory instructions in x86, where all non-memory accessing instructions are scanned from all assembly files and that list is randomly picked from. 

I have tried scanning the existing x86 assembly files for memory-accessing instructions and after some thought, decided to do the template way. The reason is that for memory accessing instructions, we only allow certain addressing parameter combinations so that it would fit into the 256 byte buffer we allocated on the stack. Not all instructions scanned from the assembly satisfy this constraint (for example, sometimes they go beyond 256 bytes). In addition, one has to be able to automatically generate those auxiliary instructions for the scanned instructions, which could be nontrivial. 

In terms of what we can claim about what is tested, I think the second approach has a stronger argument in that the instructions tested are automatically collected, whereas the first approach depends on the instruction programmer to manually add a test harness for all added instruction. If one assumes that each newly introduced instruction gets a test manually added, then the first approach has the same coverage as the second.

aqjune-aws: The Arm cosimulator also checks whether instructions in the binaries also belong to at least one of the instruction patterns too. :) `check_insns ()` of `arm/proofs/simulator_iclasses.ml` does that. Since the instruction bitpatterns listed in the first approach are slightly more general than what instructions are actually used, the second approach cannot see the instructions that will appear in the future, but well that might be okay as long as everything works fine...

Anyway, what is important I think is that we need to record this at somewhere. I will open a github issue that mentions this.

x86/proofs/x86-mem-insns.ml
pennyannn: Got it. I moved the code in x86-mem-insts.ml to simulator.ml

x86/proofs/simulator.ml
pennyannn: Added. Check new commit.

aqjune-aws (approved):

x86/proofs/x86-mem-insns.ml
pennyannn: Issue created at https://github.com/awslabs/s2n-bignum/issues/199

jargh:
Many thanks for this, Yan, which really closes one of the outstanding gaps in our soundness story! And thanks June for all the help with refining and perfecting this PR.

I agree that we need in the fairly near future to make absolutely sure we are running memory operand cosimulation on a fully representative sample of instructions, something that applies equally to aarch64. So it's good to see it logged as an issue.

Anyway, I'm now very happy with this update and will squash and merge it.

PR #190, opened 2025-02-12
Add more examples for equivalence checking, refactor equiv tactics

aqjune-aws:
This patch

- Adds more examples for equivalence checking (both x86 and arm)
- Refactor equivalence checking tactics so that Arm and x86 share more codes. :)

This patch also updates `X86_UNDEFINED_GEN_TAC` to globally assign a unique variable because otherwise this was causing some uninterpretable error when running equivalence checking tactics. :( This is also better in terms of speed because it avoids scanning the assumptions.

*Issue #, if available:*

*Description of changes:*


jargh (approved):
All looks very good, thanks - I will merge.

PR #203, opened 2025-03-14
Enable x86 indirect branch tracking in code and proofs

jargh:
*Issue #, if available:*

*Description of changes:*


pennyannn:
LGTM. I just have one question. Naturally one would expect that since ENDBR64 is a NO-OP, the existing proofs should work smoothly, because symbolic simulation will result in the same terms as if ENDBR64 is not inserted. However this PR updates the proof in the following way:
1. Trim off ENDBR64 instruction from the program and run the NOIBT proof 
2. Add the ENDBR64 instruction back and reuse the NOIBT theorem to prove the full one
What is the reason to do it this way?

x86/curve25519/bignum_add_p25519.S
```diff
@@ -40,6 +40,7 @@
 #define c3short edx
 
 S2N_BN_SYMBOL(bignum_add_p25519):
+	_CET_ENDBR
```
aqjune-aws, line 43: It seems this is starting from a tab space, which is different from other lines that start with normal spaces. Do you want to change it so that it looks prettier when opened with an editor whose tab size is non-8..?

aqjune-aws (approved):
I checked the following things:
* `specifications.txt` does not have a missing spec; for every spec `s_SUBROUTINE_CORRECT`, `s_(NOIBT_)?(WINDOWS_)?SUBROUTINE_CORRECT` exists.
* Also, it does not have a missing spec compared to the latest commit of awslabs/s2n-bignum (and also checked one name change in `BIGNUM_TRIPLE_SM2ALT_SUBROUTINE_CORRECT` :) )
* Randomly picked a few .S files, compiled them, disassembled them & checked that the new instruction appears.
* When compiled with `-DNO_IBT`, the new instruction does not appear from `objdump`.

aqjune-aws:
@pennyannn 
> I just have one question. Naturally one would expect that since ENDBR64 is a NO-OP, the existing proofs should work smoothly, because symbolic simulation will result in the same terms as if ENDBR64 is not inserted. However this PR updates the proof in the following way:

I guess it was because adding one instruction in the beginning of the functions shifted the pc offsets as well as the number of steps that are given to the STEPS tacs, which could actually break existing proofs. :)

x86/curve25519/bignum_add_p25519.S
jargh: Good point, yes, I will expand them all. I'm more or less trying to maintain the invariant of no tabs and no trailing spaces (hence no invisible differences) in s2n-bignum files, though I have occasionally compromised on tabs with Makefiles when all the backslash continuations get a bit too gruesome :-)

jargh:
I'll also address Yan's question, which is indeed a good one and reflects the evolution of my approach :-) Actually my first attempt was to modify all the proofs directly for `ENDBR64`, which I thought might be relatively painless since a lot of them use the hypothesis `bytes_loaded (word pc) (BUTLAST machinecode)`, stripping off the final return so we can also use it to talk about the Windows ABI form where there is an epilogue inserted before that return. I hoped just using `TRIM_LIST(4,1)` instead of `BUTLAST`, much the same things would work. But it turned out complicated, mainly because of the Windows ABI forms. Then there was the additional problem that some s2n-bignum routines call others via direct branches, and it didn't seem appropriate to add internal ENDBR64 instructions, which works against the ROP-protecting purpose a bit. So I thought I'd try the approach of maintaining the old proofs for the trimmed form and adding a wrapper. It was *much* easier, and retrospectively I think it's defensible to have different versions of the proof for each possible combination of the build options `-DWINDOWS_ABI` and `-DNO_IBT.`

jargh:
I've pushed some updates to
* expand the detail in the comment on ENDBR64 in line with Nevine's suggestion
* replace tabs by spaces (and remove trailing spaces) in line with June's suggestion

I made the whitespace changes in the s2n-bignum files generally except for Makefiles and tools. Almost all the changes are invisible with `git diff -b`.

PR #204, opened 2025-03-15
Add support for UMLAL2, UMLSL{2}, SMULL{2}, SMLAL{2}, SMLSL{2}

### LINKED ISSUE #202: Add support for SMULL, SMULL2, SMLAL, SMLAL2
Those instructions are needed to verify the ML-KEM AArch64 base multiplication from mlkem-native.

hanno-becker:
* Resolves #202 

This commit extends the AArch64 architecture model with support for `UMLAL2, UMLSL{2}, SMULL{2}, SMLAL{2}, SMLSL{2}`.

This follows the existing templates for UMULL and UMLAL. The difference in encoding between Sxxx and Uxxx is bit-29 only, which is 1 for unsigned and 0 for signed variants. Similarly, xMLAL and xMLSL only differ in bit 13.

Semantically, we merely need to switch from zero-extend to sign-extend, and replace word_add by word_sub for xMLSL variants.


hanno-becker:
@jargh @aqjune I don't seem to have the power to request reviews -- could one of you have a look, please?

jargh:
I think I've volunteered myself - thanks!

jargh:
Your comment on how closely analogous these are to other instruction models reminds me that eventually it would be nice to do a refactoring of some of these common patterns into generic code. But that's something for another day, which I think will only slow us down at this point - the presence of multiple type instances of core operations also makes doing so a bit fiddly. Anyway I'll take a closer look at the PR and approve it if I don't see any issues, hopefully by today.

jargh:
I have looked over this and everything looks good, though I must admit I didn't check every small detail. I did however run the cosimulator for 5 hours on a 192-core Graviton4 restricted just to the newly added instructions and all was well, which together with the general look of rightness I find pretty convincing.

jargh (approved):
I'm happy to merge this if you like. The only small suggestion I have (but I think it may apply more widely than your newly added instructions, and I don't regard it as critical) is that I prefer also to cosimulate some of the illegal combinations to make sure they are correctly *not* accepted, which in this case would mean allowing any two bits in the size field like this (oh, and without the comment about the size not being 11)
```

  (*** SMLAL ***)
  "000011100xxxxxxx100000xxxxxxxxxx"; (* src: .b, .h *)

  (*** SMLAL2 ***)
  "010011100xxxxxxx100000xxxxxxxxxx"; (* src: .b, .h *)

  (*** SMLSL ***)
  "000011100xxxxxxx101000xxxxxxxxxx"; (* src: .b, .h *)

  (*** SMLSL2 ***)
  "010011100xxxxxxx101000xxxxxxxxxx"; (* src: .b, .h *)

  (*** SMULL ***)
  "000011100xxxxxxx110000xxxxxxxxxx"; (* size!=11 *)

  (*** SMULL2 ***)
  "010011100xxxxxxx110000xxxxxxxxxx"; (* size!=11 *)

  (*** UMLAL2 ***)
  "011011100xxxxxxx100000xxxxxxxxxx"; (* src: .b, .h *)

  (*** UMLSL ***)
  "001011100xxxxxxx101000xxxxxxxxxx"; (* src: .b, .h *)

  (*** UMLSL2 ***)
  "011011100xxxxxxx101000xxxxxxxxxx"; (* src: .b, .h *)

```

hanno-becker:
@jargh Thank you for your review!

I agree with your point, but it affects all Neon multiplication instructions, so it would make sense to handle all of them at once. One should also look for other constraints that could be negatively tested in a similar way.

hanno-becker:
@jargh I opened a separate draft-PR https://github.com/awslabs/s2n-bignum/pull/205 for this.

hanno-becker:
> Your comment on how closely analogous these are to other instruction models reminds me that eventually it would be nice to do a refactoring of some of these common patterns into generic code. But that's something for another day, which I think will only slow us down at this point - the presence of multiple type instances of core operations also makes doing so a bit fiddly.

I opened #206 to track this.

jargh:
This is great, I'll merge in this one then.

PR #205, opened 2025-03-17
AArch64/SIMD: Add negative decoding tests for invalid sizes

hanno-becker:
* ~Based on #204~

Various NEON instructions have a 2-bit size parameter but are
only valid for the values {0b00, 0b01, 0b10}. Notably, this
is the case for all SIMD integer multiplication instructions.

For those instructions, `simulator_iclasses.ml` previously
listed only the valid encodings. This would test that the functional
description of valid encodings was correct, but it would not test
that the architecture and the model agreed on them being invalid.

This commit relaxes the encoding patterns in `simulator_iclasses.ml`
so that invalid size encodings are considered as well.

A similar change may be possible for immediates required to be
non-zero, but the architecture is not clear on what decoding does
with them: It says `SEE(asimdimm)` rather than `UNDEF`. Those
are therefore not touched for now.


hanno-becker:
@jargh I opened a separate draft-PR https://github.com/awslabs/s2n-bignum/pull/205 for this.

jargh:
Yes, this is the sort of thing I have in mind. Generally, the instructions understood by the s2n-bignum decoder and executed by the symbolic simulation machinery are expected to be a proper subset of those accepted on the hardware, because many instructions are missing (though more are being added all the time). I suppose in principle we'd like to check all clearly illegal instruction encodings to make sure the decoder/simulator doesn't accept them. In practice, I had the more modest goal of just keeping bit patterns for instruction forms simple and general where possible, considering it a positive rather than negative feature that it executes a sprinkling of illegal cases.

hanno-becker:
@jargh Ok, thank you! Let me know if you'd like any further changes, or feel free of course to just push them atop.

aqjune-aws (approved):
Do you want to rebase your change to the latest version of the main branch? Or I can update this PR as well.

hanno-becker:
Thank you @aqjune-aws. Are we good to merge?

jargh:
Looks very good to me too - even the comments are right! I'll make a few last simulator runs as a final sanity check and merge later today.

PR #207, opened 2025-03-17
Add read-only section support for ELF AArch64

aqjune-aws:
This patch adds support for reading the read-only section of ELF (AArch64) and using it.

This is achieved by extending the ELF loader to
(1) find the .rodata section and read it
(2) read the symbol table and find the objects that are defined in `.rodata` (3) read the relocation info to record which instructions must be updated when loading it.

For ELF AArch64, GCC is typically using `adrp` + `add` combination (which is also equivalent to the `adrl` pseudoinstruction) to add the PC-relative offset and correctly find the symbol address.

When relocation is involved, the machine-code isn't just a list of bytes, but in fact a function that takes the virtual addresses of the symbols and returns the correct, 'patched' machine code (which is a byte list). `define_assert_relocs_from_elf` returns that, as well as a list of new definitions describing the contents of constant data.

This patch also includes `arm/tutorial/symtab_rodata.ml`, which is an example that loads constant data.

I found that the following references were helpful:
- https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#5733relocation-operations for overall meanings of AArch64-specific relocations
- https://refspecs.linuxbase.org/elf/gabi4+/ for the structure of an ELF file
- https://github.com/lattera/glibc/blob/master/elf/elf.h#L2731 for kinds of relocations

*Issue #, if available:*

*Description of changes:*


jargh (commented):
This is now all working very well, thank you! I was able to update the `curve25519_x25519base` code to put its constant tables in rodata and then prove it using the new infrastructure, with relatively few changes from the original proof :-)

PR #208, opened 2025-03-17
Add AESNI Instructions for x86_64

pennyannn:
This PR does the following:

1. Adds infrastructure for supporting SSE/AESNI instructions. Current XMMs and YMMs register definitions support writing zeros to upper half of YMMs and ZMMs respectively when being written to. SSE and AESNI instructions don't write zeros but will keep the upper half values (See https://www.felixcloutier.com/x86/aesenc as an example). This PR defines XMM_SSE and YMM_SSE versions of these registers that don't write zeros. Internally, the symbolic simulation will reduce reads of YMM_SSE into reads of YMM, creating a theorem statement about YMMs, which existing proof framework could work with.
2. Add five AESNI instructions: AESENC, AESENCLAST, AESDEC, AESDECLAST and AESKEYGENASSIST. These instructions are currently manually tested by hand-written tests. The hand-written tests could be removed once the AES-XTS assembly program is integrated.
3. Refactors common AES functionalities into common/aes.ml, which arm/proofs/aes.ml and x86/proofs/ase.ml both depend on.
4. Added a couple lemmas that reason about `word_subword` and `word_zx`. I'm not sure if they should be moved elsewhere, like into common/misc? HOL Light?


x86/proofs/x86.ml
```diff
@@ -310,6 +310,139 @@ add_component_alias_thms
...
+    SUBGOAL_THEN `((word_subword (x:N word) (0,len)):M word) =
+      ((word_zx ((word_subword x (0,dimindex(:N))):N word)):M word)`
+    ASSUME_TAC THENL
+    [MATCH_MP_TAC(ISPECL [`x:N word`;`0:num`;`(dimindex(:N):num)`;`len:num`]
+              WORD_SUBWORD_EQUAL_SUBWORD_THEN_ZX) THEN
+     ASM_REWRITE_TAC[] THEN
+     ASM_ARITH_TAC THEN
+     ASM_REWRITE_TAC[WORD_SUBWORD_N_EQUAL]; ALL_TAC] THEN
+    ASM_REWRITE_TAC[WORD_SUBWORD_N_EQUAL]
+    );;
```
aqjune-aws, line 370: Do you want to move these theorems to common/misc.ml ? misc.ml will have a cluster of word related lemmas.

x86/proofs/x86.ml
```diff
@@ -310,6 +310,139 @@ add_component_alias_thms
...
+  (`!s:x86state. read YMM13_SSE s = read YMM13 s`,
+   READ_YMM_SSE_TAC YMM13_SSE YMM13);;
+
+let READ_YMM14_SSE_EQUIV = prove
+  (`!s:x86state. read YMM14_SSE s = read YMM14 s`,
+   READ_YMM_SSE_TAC YMM14_SSE YMM14);;
+
+let READ_YMM15_SSE_EQUIV = prove
+  (`!s:x86state. read YMM15_SSE s = read YMM15 s`,
+   READ_YMM_SSE_TAC YMM15_SSE YMM15);;
```
aqjune-aws, line 444: Can we make these 16 theorems compact using `map`? For example,
```
let READ_YMM_SSE_EQUIV:thm list = map (fun (the_goal,ssereg,reg) -> prove(t, READ_YMM_SSE_TAC ssereg reg))
    [(`!s:x86state. read YMM0_SSE s = read YMM0 s`,`YMM0_SSE`,`YMM0`);
     (`!s:x86state. read YMM1_SSE s = read YMM1 s`,`YMM1_SSE`,`YMM1`);
     ...]
```

x86/proofs/x86.ml
```diff
@@ -310,6 +310,139 @@ add_component_alias_thms
  [YMM0; YMM1; YMM2; YMM3; YMM4; YMM5; YMM6; YMM7;
   YMM8; YMM9; YMM10; YMM11; YMM12; YMM13; YMM14; YMM15];;
```
aqjune-aws, line 312: Should we leave a proper comment that explains when these registers are to be used (something like line 243)?

Probably moving your new definitions (YMM_SSE, XMM_SSE) to line 501 (line 345 before changes) would make the contents of the new comment clearer.

x86/proofs/x86.ml
pennyannn: Good idea. Moved to common/misc.ml. I have also tested that `WORD_SUBWORD_EQUAL_SUBWORD_THEN_ZX` and `WORD_SUBWORD_N_EQUAL` don't depend on anything in s2n-bignum, so I named the section as "More word lemmas".

x86/proofs/x86.ml
pennyannn: Thank you for the suggestion! Done.

x86/proofs/aes.ml
```diff
@@ -0,0 +1,123 @@
...
+         Comb(Const("word",_),roundkey))
+    when is_numeral state && is_numeral roundkey -> AESDECLAST_HELPER_CONV tm
+    | _ -> failwith "AESDECLAST_REDUCE_CONV: inapplicable";;
+let AESKEYGENASSIST_REDUCE_CONV tm =
+  match tm with
+    Comb(Comb(Const("aeskeygenassist",_),
+         Comb(Const("word",_),x)),
+         Comb(Const("word",_),imm8))
+    when is_numeral x && is_numeral imm8 -> AESKEYGENASSIST_HELPER_CONV tm
+    | _ -> failwith "AESKEYGENASSIST_HELPER_CONV: inapplicable";;
```
aqjune-aws, line 123: Do you want to add examples that tests the definitions of `aesenclast`, `aesdec`, using the input and outputs in https://www.intel.com/content/dam/doc/white-paper/advanced-encryption-standard-new-instructions-set-paper.pdf ? For example, `aesenclast` can be tested using Figure 12. AESENCLAST Example in the pdf.
Other than this, this looks all good to me! Thanks for your hard work & sorry for delay.

x86/proofs/aes.ml
```diff
@@ -0,0 +1,123 @@
+(*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT-0
+ *)
```
aqjune-aws, line 4: Oops, one missing comment that wasn't submitted.. What do you think about moving `needs common/aes.ml` to here from x86/proofs/base.ml ? This will help the dependency structure clearer. Same can be done for arm.

x86/proofs/aes.ml
aqjune-aws: Not widespread, but I have a few tests using OCaml's assert here: https://github.com/awslabs/s2n-bignum/blob/main/common/equiv.ml#L59-L65
There was SHA256 formalization added by an intern but they did not have the test vectors.
Or probably we can rely on `prove` in this case, something like: ``prove(`aesdec ... = ..`, CONV_TAC (LAND_CONV AESDEC_REDUCE_CONV) THEN CONV_TAC WORD_REDUCE_CONV)`` (with minor modifications in tactic if necessary).

x86/proofs/aes.ml
```diff
@@ -0,0 +1,155 @@
...
+needs "common/aes.ml";;
+
+(* References:
+   Intel 64 and IA-32 Architectures Software Developer's Manual Volume 1:
+     Basic Architecture, page 311-316
```
aqjune-aws, line 9: Leaving a note: this links to "Vol 1. 12-19"

pennyannn, line 9: Page number should be `Vol. 1 12-19`.

pennyannn: Duplicated.

x86/proofs/simulator.ml
```diff
@@ -257,6 +257,22 @@ let iclasses = iclasses @
  [0x66; 0x87; 0xfe]; (* XCHG (% si) (%di) *)
  [0x66; 0x87; 0xfe]; (* XCHG (% si) (%di) *)
  [0x40; 0x86; 0xfe]; (* XCHG (% sil) (%dil) *)
+ (* TODO: remove these tests once the AES-XTS program is integrated *)
```
aqjune-aws, line 260: Leaving a note (not to forget): in the future, if AES-XTS program is integrated the test/benchmark will need to check CPUID.01H:ECX.AESNI[bit 25] = 1 . (related: https://stackoverflow.com/questions/50609654/how-to-check-aes-ni-support-with-cpuid)

PR #210, opened 2025-03-21
Add BIC immediate, CMHI, CNT, single-precision FMOV and UADDLV to ARM model

jargh:
*Issue #, if available:*

*Description of changes:*


arm/proofs/decode.ml
```diff
@@ -406,13 +423,22 @@ let decode = new_definition `!w:int32. decode w =
   | [0:1; q; 1:1; 0b011110:6; immh:4; abc:3; cmode:4; 0b01:2; defgh:5; Rd:5] ->
     // MOVI, USHR (Vector), USRA (Vector), SLI (Vector), SRI (vector)
     if val immh = 0 then
-      // MOVI
-      if q then
+      // MOVI, 128-bit only
+      if cmode = word 0b1110 /\ q then
```
aqjune-aws, line 427: This change looks fine to me. This additional guard is necessary because `arm_adv_simd_expand_imm` can now handle more cases.

pennyannn (commented):
I have done the following things:
1. Checked the modeling against ARM manual 
2. Running this PR on my graviton3 machine and I found all modeled instruction variants in the output

jargh:
Thanks for the eagle-eyed review! I think the new commit fixes all that. I took the chance to add other new instructions that were added recently to the `allowed_asm` list, even though many of them are not yet used in the code, or used only in some branches.

aqjune-aws (approved):
- Checked that the bit formats are equal to what the Arm doc describes.
- Also checked that cosimulator does not fail for > 10 minutes, after excluding all scalar instructions to focus on possibly affected vector instructions.

aqjune-aws:
Probably we can merge after Nevine's patch is merged first, since her PR is reviewed first..?

Issue #199, opened 2025-03-06
Check that cosimulation testing of memory-accessing instructions cover all instructions used in assembly


pennyannn:
Cosimulation testing of memory-accessing instructions in x86_64 and aarch64 are done by randomly-generated instructions based on templates. To ensure these randomly generated instructions cover all cases of instructions in the assembly code, additional check needs to be implemented. Similar mechanism could be found at `check_insns ()` of `arm/proofs/simulator_iclasses.ml`.

PR #211, opened 2025-03-21
Add Arm instructions needed for AES-XTS

nebeid:
*Description of changes:*
Added decoding for:
- NOP
- LD1/ST1 post-index register offset and no offset for both 64-bit and 128-bit registers
- LD1/ST1 2 128-bit registers, post-index immediate offset
- LDRB/STRB with (shifted) register index, no shift.

### DIFF
--- a/arm/proofs/simulator_iclasses.ml
+++ b/arm/proofs/simulator_iclasses.ml
@@ -472,7 +472,7 @@ let check_insns () =
     "0x001100010000000111xxxxxxxxxxxx";
 
     (*** st1 (1 register, no Post-immediate offset) ***)
-    "0x001100100111110111xxxxxxxxxxxx";
+    "0x001100000000000111xxxxxxxxxxxx";
 
     (*** ld1 (2 registers, Post-immediate offset) 128-bit ***)
     "01001100110111111010xxxxxxxxxxxx";
--- a/arm/proofs/decode.ml
+++ b/arm/proofs/decode.ml
@@ -342,20 +342,22 @@ let decode = new_definition `!w:int32. decode w =
   // that reuses functions of LDR/STR for LD1/ST1 is okay.
 
   // datasize = 64
+  // Post-immediate offset and post-register offset
   | [0:1; 0:1; 0b0011001:7; is_ld; 0:1; Rm:5; 0b0111:4; size:2; Rn:5; Rt:5] ->
     SOME (arm_ldst_d is_ld Rt (XREG_SP Rn)
     (if val Rm = 31 then (Postimmediate_Offset (word 8))
                     else Postreg_Offset (XREG' Rm)))
-  // no Postimmediate_Offset
+  // No offset
   | [0:1; 0:1; 0b0011000:7; is_ld; 0b000000:6; 0b0111:4; size:2; Rn:5; Rt:5] ->
     SOME (arm_ldst_d is_ld Rt (XREG_SP Rn) No_Offset)
 
   // datasize = 128
+  // Post-immediate offset and post-register offset
   | [0:1; 1:1; 0b0011001:7; is_ld; 0:1; Rm:5; 0b0111:4; size:2; Rn:5; Rt:5] ->
     SOME (arm_ldst_q is_ld Rt (XREG_SP Rn)
       (if val Rm = 31 then (Postimmediate_Offset (word 16))
                       else Postreg_Offset (XREG' Rm)))
-  // no Postimmediate_Offset
+  // No offset
   | [0:1; 1:1; 0b0011000:7; is_ld; 0b000000:6; 0b0111:4; size:2; Rn:5; Rt:5] ->
     SOME (arm_ldst_q is_ld Rt (XREG_SP Rn) No_Offset)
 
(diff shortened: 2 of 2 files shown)


jargh:
I've taken a first look and I think all the decoding part and the logic of the instructions look good and nicely generalize what we have. There is however one issue, which is that the `Register_Offset` field is only used for an internal offset from the base register in the read/write address calculation `[rn, #offset]`. To accommodate the post-immediate register offset that modifies the register we will need to extend the type of offsets. You can see that there is a comment that we're currently explicitly ignoring this case and only doing pre/post immediates:
https://github.com/awslabs/s2n-bignum/blob/main/arm/proofs/instruction.ml#L687-L688
(So the current definition would in effect give `[rn + rm]` rather than `[rn], rm` as desired.)
This, unfortunately, requires a little bit of modification of some of these foundational definitions (for example, if we add the new field, `offset_writeback` will now need to take the state as an additional argument so it can read the register value). If you feel up to trying it, I'm very happy to let you, but also I would be happy to start with your PR and make some tweaks myself.

nebeid:
> I've taken a first look and I think all the decoding part and the logic of the instructions look good and nicely generalize what we have. There is however one issue, which is that the `Register_Offset` field is only used for an internal offset from the base register in the read/write address calculation `[rn, #offset]`. To accommodate the post-immediate register offset that modifies the register we will need to extend the type of offsets. 

Thank you, jargh, for taking a close look and for the details in how to fix it. I attempted to add the correct mode.

jargh:
The semantics all look great to me now, thanks! I have been writing some extensions of the cosimulator to cross-validate these additions, and in the process uncovered a couple of extra tweaks that help the simulation to go through smoothly, basically just propagating things about your newly used offset versions into the actual simulation machinery. Not only is `Postreg_Offset` new, but it turns out that `No_Offset`, although present, had never been used either and was not properly supported in the simulation simplifications, which explains the addition there. The attached patch will add these two little things.
[offset.patch](https://github.com/user-attachments/files/19510121/offset.patch)

jargh:
Here is a larger patch with one more little tweak and the additional cosimulation harnesses. So far, everything checks out fine, as far as I can see, so I'd be happy to merge this version in.
[morecosim.patch](https://github.com/user-attachments/files/19514284/morecosim.patch)

jargh:
I think I've rolled in the patch now.

arm/proofs/simulator_iclasses.ml
```diff
@@ -459,6 +462,24 @@ let check_insns () =
...
+    "0x001100110xxxxx0111xxxxxxxxxxxx";
+
+    (*** st1 (1 register, Post-register offset) ***)
+    "0x001100100xxxxx0111xxxxxxxxxxxx";
+
+    (*** ld1 (1 register, no Post-immediate offset) ***)
+    "0x001100010000000111xxxxxxxxxxxx";
+
+    (*** st1 (1 register, no Post-immediate offset) ***)
+    "0x001100100111110111xxxxxxxxxxxx";
```
pennyannn, line 475: Does "no Post-immediate offset" mean "no offset"? I checked the manual and this pattern seems different: https://developer.arm.com/documentation/ddi0602/2023-12/SIMD-FP-Instructions/ST1--multiple-structures---Store-multiple-single-element-structures-from-one--two--three--or-four-registers-?lang=en

pennyannn (commented):
LGTM except the one issue.
I did the following:
1. Checked that the bit pattern in iclasses covers the right instructions
2. Ran the cosimulation test on my Graviton3 machine and checked that NOP appeared in the test. There is an issue that since LD1/ST1 is dispatched to LDR/STR and LDP/STP, one cannot easily figure out if these instructions are tested. Maybe a solution is to add a layer of function called `arm_LD1/ST1`. But this could be done later.

arm/proofs/simulator_iclasses.ml
nebeid: Thanks, Yan.
- yes, "no Post-immediate offset" is equivalent to "no offset" in the documentation. I will change the comments here and in decode.ml
- you're right, I may have copy-pasted without changing. The bit pattern should be the same as the ld1 above it except for the load (L) bit being set to 0. I'll fix it.

I have a question, should these now be moved out of the `skipping_iclasses`?

pennyannn: I think they should be kept in there since it belongs to a check for eliminating testing of LD/ST instructions from the assembly.

pennyannn (approved):

arm/proofs/simulator_iclasses.ml
jargh: Thank you both for the last checks and improvements! I agree with keeping things in `skipping_iclasses` for now, until we finally have a comprehensive check correlating our cosimulation of memory instructions with the patterns. We have an open issue for this so that we can't forget: https://github.com/awslabs/s2n-bignum/issues/199

jargh: The question of having separate `LD1`/`ST1` `arm_xxx` instructions is an interesting one, which we might want to consider in the future (and we would definitely need to if we ever model big-endian ARM). If nothing else it could simply be an abbreviation to maintain a better correspondence with the usual assembler syntax, rather as `No_Offset` is defined as a standalone concept but equal semantically to an immediate zero offset.

jargh:
Should we remove the BUG label now? Assuming that the final CI checks work I plan to merge this.

PR #212, opened 2025-03-24
Add MOVAPS, MOVUPS, MOVDQA, MOVDQU

pennyannn:
*Description of changes:*

This PR adds SSE/SSE2 instructions `MOVAPS`, `MOVUPS`, `MOVDQA`, `MOVDQU` for AES-XTS on x86_64. Below are the detailed changes made:
1. `MOVAPS` and `MOVDQA` are 16-bytes aligned. To reason about alignment, the conversions and tactics could be reused from Arm. Therefore I copied related code from `arm/proofs/arm.ml` and created a new file in `common` called `alignment.ml` to make both `arm/proofs/base.ml` and `x86/proofs/base.ml` load it.
2. Added several harnesses for cosimulation testing of the four MOVxxx instructions. 
3. The `x86_CONV` conversion function is updated to allow handling of alignment related transformations. 
4. Added an assumption `additional_assumption` to `template`. When alignment is required for an instruction, `additional_assumption` is substituted with `align 16 stackpointer`; otherwise, it is `T`.
[OLD_VERSION: Added a precondition `aligned 16 stackpointer` to definition of `template` in x86/proofs/simulator.ml. This is a valid addition because Intel x86_86 stackpointer is 16-bytes aligned. See [Intel® 64 and IA-32 Architectures Software Developer's Manual Volume 1: Basic Architecture](https://cdrdv2.intel.com/v1/dl/getContent/671436), 6.2.2. stack alignment.]
5. Minor refactoring.

### DIFF
--- a/x86/proofs/x86.ml
+++ b/x86/proofs/x86.ml
@@ -861,15 +861,25 @@ let x86_LZCNT = new_definition
          ZF := (val z = 0) ,,
          UNDEFINED_VALUES[OF;SF;PF;AF]) s`;;
 
-(* Only deal with register-register exchange *)
-let x86_XCHG = new_definition
- `x86_XCHG dest src s =
-    let temp = read dest s in
-    (dest := read src s ,, src := temp) s`;;
-
 let x86_MOV = new_definition
  `x86_MOV dest src s =
         let x = read src s in (dest := x) s`;;
+
+let x86_MOVAPS = new_definition
+ `x86_MOVAPS dest src s =
+    let x = read src s in (dest := x) s`;;
+
+let x86_MOVDQA = new_definition
+ `x86_MOVDQA dest src s =
+    let x = read src s in (dest := x) s`;;
+
+let x86_MOVDQU = new_definition
+`x86_MOVDQU dest src s =
+   let x = read src s in (dest := x) s`;;
+
+let x86_MOVUPS = new_definition
+ `x86_MOVUPS dest src s =
+    let x = read src s in (dest := x) s`;;
 
 (*** These are rare cases with distinct source and destination
  *** operand sizes. There is a 32-bit to 64-bit version of MOVSX(D),
@@ -1261,6 +1271,12 @@ let x86_VPXOR = new_definition
         let z = word_xor x y in
         (dest := (z:N word)) s`;;
 
+(* Only deal with register-register exchange *)
+let x86_XCHG = new_definition
+ `x86_XCHG dest src s =
+    let temp = read dest s in
+    (dest := read src s ,, src := temp) s`;;
+
 let x86_XOR = new_definition
  `x86_XOR dest src s =
         let x = read dest s and y = read src s in
@@ -1404,6 +1420,10 @@ let OPERAND8 = define
   OPERAND8 (Memop w ea) s =
        (if w = Byte then memory :> bytes8 (bsid_semantics ea s)
         else ARB)`;;
+
+let aligned_OPERAND128 = define
+ `(aligned_OPERAND128 (Simdregister r) s <=> T) /\
+  (aligned_OPERAND128 (Memop w ea) s <=> aligned 16 (bsid_semantics ea s))`;;
 
 (* ------------------------------------------------------------------------- *)
 (* Stating assumptions about instruction decoding                            *)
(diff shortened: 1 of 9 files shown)


x86/proofs/x86.ml
```diff
@@ -2746,10 +2777,11 @@ let X86_CONV (decode_ths:thm option array) ths tm =
      GEN_REWRITE_CONV (RAND_CONV o TOP_DEPTH_CONV) [GSYM ADD_ASSOC] THENC
      RAND_CONV NUM_REDUCE_CONV) THENC
    TOP_DEPTH_CONV COMPONENT_WRITE_OVER_WRITE_CONV THENC
-   GEN_REWRITE_CONV (SUB_COMPONENTS_CONV o TOP_DEPTH_CONV) ths THENC
+   GEN_REWRITE_CONV TOP_DEPTH_CONV ths THENC
```
pennyannn, line 2780: Question: I made this rewrite rule more general because there is an if-condition for MOVAPS/MOVDQA which checks alignment. Rewrites using the assumptions is required for simplifying that if-condition. However, the original version is more constrained in that it only concerns internals of `write` and `read` expressions. I'm wondering if this change is okay.

pennyannn: Okay, this change is indeed a bad one as shown in the CI run that failed the x86 proofs build. With a second thought, I think it is the best to put `ALIGNED_16_CONV` in `X86_FORCE_CONDITIONAL_CONV`, since this conversion is used for reducing if-conditions. See [commit](https://github.com/awslabs/s2n-bignum/pull/212/commits/8324ae0cf95481af70298c38076dd63fed136fb9).

aqjune-aws:
Hi Yan,

> Added a precondition aligned 16 stackpointer to definition of template in x86/proofs/simulator.ml. This is a valid addition because Intel x86_86 stackpointer is 16-bytes aligned. See [Intel® 64 and IA-32 Architectures Software Developer's Manual Volume 1: Basic Architecture](https://cdrdv2.intel.com/v1/dl/getContent/671436), 6.2.2. stack alignment.

I looked at 6.2.2, and it seems the paragraph is saying 16 bits instead of 16 bytes:

> The stack pointer for a stack segment should be aligned on 16-bit (word) or 32-bit (double-word) boundaries,
depending on the width of the stack segment.

But from a few online web sites I am seeing that stack pointers are supposed to be 16 bytes aligned, but only for particular ABIs. I found this https://learn.microsoft.com/en-us/cpp/build/stack-usage?view=msvc-170#function-types .
Should we reconsider having the 16-bytes aligned stack pointer as a precondition in the simulator..?

pennyannn:
> But from a few online web sites I am seeing that stack pointers are supposed to be 16 bytes aligned, but only for particular ABIs. I found this https://learn.microsoft.com/en-us/cpp/build/stack-usage?view=msvc-170#function-types . Should we reconsider having the 16-bytes aligned stack pointer as a precondition in the simulator..?
Ah, nice catch. Sorry for the misread. 
Yes, I think another solution is to use different proof `template` for instructions that should be aligned memory and ones that don't require this. How do you like this idea?

aqjune-aws:
I think that for the AVX2 MOVs the aligned stack pointer makes sense; wondering whether this additional alignment assumption might affect other normal memory loading instructions in x86. I guess a normal mov instruction does not assume that the pointer is aligned, no?

pennyannn:
Yes. What you said is correct. I'm just remembering this observation that our actual machine simulation run probably obey the 16-byte alignment rule. 
I will update the PR with changes like we discussed. :)

pennyannn:
Refactored cosimulation code to only add alignment assumption when needed. See [new commit](https://github.com/awslabs/s2n-bignum/pull/212/commits/fc06ae5f144ae59a730888fcd00165ae402f636d).

aqjune-aws (approved):
- The decoder worked reasonably for my hand-wrote assembly instructions for MOVAPS/MOVUPS/MOVDQA/MOVDQU whose operands are (1) register-registers, (2) register-memory with various memory operand forms, and (3) memory-register with ditto.
- Running cosimulator for 30 mins worked fine. (1) memory instructions are left only (since they are affected by the new flag) and (2) the four new instructions left only
- The described non-VEX cases of these four instructions looked fine.

I have a concern about the increasing duplicatedness in the x86 simulator's memory instruction codes, btw. Do we want to collect the shared parts between the `cosimulate_*` functions and make them more compact? Probably we can add more comments too. This can be a follow-up patch.
Another improvements could be (1) simulating the aligned instructions when the memory address was not actually aligned, and (2) adding register-to-register copy cases. I think the simulator will already be able to deal with (1), since it can handle crashing instructions (but it might not). I think this can be all included in x86 simulator refactoring patch.

Issue #146, opened 2024-08-28
A cautionary tale of semicolons in assembly


ctz:
In arm/Makefile there is:

https://github.com/awslabs/s2n-bignum/blob/08bf556adff16d9afb9952474442cb8c9b94b9aa/arm/Makefile#L36-L39

On macos `as` _treats semicolons as beginning of a line comment_.[^1] That means that this build step is very, very load bearing, and if downstream users merely take the assembly as-is, they will have a bad time. For example:

```
#define movbig(nn,n3,n2,n1,n0)                                      \
        movz    nn, n0;                                             \
        movk    nn, n1, lsl #16;                                    \
        movk    nn, n2, lsl #32;                                    \
        movk    nn, n3, lsl #48
```

Gets assembled as just `movz nn, n0`.

That doesn't pass even the most cursory testing, but also it's quite subtle and it took me a while to work out why aarch64-linux worked but aarch64-apple didn't.

Not sure what the ideal solution is here, but reporting the issue seems like a start.

[^1]: https://github.com/llvm/llvm-project/blob/llvmorg-12.0.0-rc3/llvm/lib/Target/AArch64/MCTargetDesc/AArch64MCAsmInfo.cpp#L41

ctz:
Going to close this as I don't think there is a feasible resolution. Maybe it will help someone in the future if they are confused in a similar way I was.

Thanks for the project!

PR #213, opened 2025-03-25
Fix Mac ARM multiline macros

justsmth:
*Issue #, if available:*
* Related: https://github.com/aws/aws-lc/pull/2287
  * Note that all relevant tests are passing.

*Description of changes:*
* On MacOS/ARM64, the assembler treats the semicolon (`;`) as the beginning of a comment. For multiline macros that use the semicolon to separate lines, every line of the macro after the first is treated as being part of a comment.
* To work around this, AWS-LC has been doing custom processing of the s2n-bignum assembly files: https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/CMakeLists.txt#L329
* This change replaces the use of semicolons with a `__LF` macro that expands to an appropriate line separator depending on the platform.

*Callout*
* Changes to the assembly files were all performed by a simple CLI command executed from the `arm` directory:
```
find . -name "*.S" | xargs sed -I '' -e 's/; \(.*\\\)$/ __LF\1/g'
```

### DIFF
--- a/arm/curve25519/bignum_inv_p25519.S
+++ b/arm/curve25519/bignum_inv_p25519.S
@@ -80,618 +80,618 @@
 // [ m10  m11]
 
 #define divstep59()                                                     \
-        and     x4, x2, #0xfffff;                                       \
-        orr     x4, x4, #0xfffffe0000000000;                            \
-        and     x5, x3, #0xfffff;                                       \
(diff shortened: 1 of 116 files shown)


ctz:
related: #146

jargh (approved):
This all looks great, thanks! I've tried it out on a few platforms and it all works smoothly.

PR #215, opened 2025-03-28
Add additional instructions required by AES-XTS on x86_64

pennyannn:
*Description of changes:*

This PR adds additional SSE/SSE2 instructions that are used in AES-XTS on x86_64, including multi-byte `NOP`, `PADDD`,`PADDQ`,`PAND`,`PCMPGTD`, `PSHUFD`,`PSRAD`, and `PXOR`. 

They are tested through hand-written tests. The tests could be removed once the AES-XTS assembly program is integrated into s2n-bignum.

In addition, this PR updates the `allowed_asm` file.

Notes:
The multi-byte `NOP` appeared in the assembled binary because of the `.align` directive. These `.align` directives are used for making sure program instructions are aligned properly (https://stackoverflow.com/questions/18113995/performance-optimisations-of-x86-64-assembly-alignment-and-branch-prediction). These `.align` directives are assembled into multi-byte `NOP`s to pad up the space. In this PR we add multi-byte NOP that supports from 2-bytes to 9-bytes. However, these multi-byte NOP can be really wild. See https://stackoverflow.com/questions/27714524/x86-multi-byte-nop-and-instruction-prefix, where it points out that Intel can have 11-byte NOP with redundant prefixes. 10-byte and up NOP requires changes to data structures and will be done in another [PR](https://github.com/awslabs/s2n-bignum/pull/218).
XORPS is subject to constant-time uncertainty because it is not part of the DOIT list. Removed and will be submitted through another PR (https://github.com/awslabs/s2n-bignum/pull/225).


x86/allowed_asm
```diff
@@ -38,6 +45,7 @@
 : cmpq$
 : dec$
 : decq$
+: div$
```
aqjune-aws, line 48: Oops, this looks suspicious, isn't it..? I think the purpose of allowed_asm was to find out non-constant-time instructions.

In this PR, should we only add the newly formalized instructions too `allowed_asm`? Maybe we can add the remaining instructions when AES-XTS is eventually checked in to s2n-bignum.

x86/proofs/x86.ml
```diff
@@ -979,6 +986,67 @@ let x86_OR = new_definition
          OF := F ,,
          UNDEFINED_VALUES[AF]) s`;;
 
+let x86_PADDD = new_definition
+  `x86_PADDD dest src s =
+    let x = read dest s in
+    let y = read src s in
+    let r0 = word_add
+      ((word_subword:(128 word->num#num->32 word)) x (0,32))
+      ((word_subword:(128 word->num#num->32 word)) y (0,32)) in
```
aqjune-aws, line 995: You can slightly shrink this to:
```
let r0 = word_add (word_subword x (0,32)) (word_subword y (0,32)):32 word in
```
:)

x86/allowed_asm
pennyannn: Noted. I have removed ones that are not constant time according to this list: https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/resources/data-operand-independent-timing-instructions.html

x86/proofs/x86.ml
pennyannn: Done this, also in a couple other functions.

aqjune-aws (approved):
Thanks for separating XORPS from this PR..!
I checked that the implementation is consistent with the descriptions of the instructions in x86 docs, and cosimulator did not found inconsistencies for the newly added instruction patterns.
I added a small patch that uses simdN/usimdN to simplify the formal semantics. Also, added EXPAND_SIMD_RULE which makes the structure of x86 semantics for SIMD analogous to that of Arm side.

PR #218, opened 2025-04-05
Add support for multi-byte NOP with segment override prefix and duplicated operand-size override prefix

pennyannn:
*Description of changes:*

In x86_64 assembly, `.align` directive gets expanded into multi-byte NOP that uses segment override prefix and duplicated operand-size override prefix. See https://stackoverflow.com/questions/27714524/x86-multi-byte-nop-and-instruction-prefix
This PR extends current x86_64 decoder to support this kind of long NOPs. 

In addition, this PR fixes a bug in the testing of PUSH/POP with RSP register. 


jargh (approved):
This all looks absolutely great! Many thanks for negotiating all that machinery around prefixes and the decoder, which is highly non-trivial...

PR #219, opened 2025-04-05
Add single-register case of TBL instruction to ARM model

jargh:
This instruction uses the bytes in one argument as indices into the bytes in the other one. Any indexing bytes >= 16 result in a zero result for that lookup, by definition, and this is implicit in the definition of the semantics using "word_subword". Note that our cosimulation testing in "simulator.ml" with random numbers will have a strong bias towards this trivial case for each lookup. However, there's around a 50% chance that at least one of the 8 or 16 lookups per instruction (for data sizes 64 or 128) will be nontrivial, so it seems that it still gives adequate coverage.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/arm/proofs/decode.ml
+++ b/arm/proofs/decode.ml
@@ -866,6 +866,11 @@ let decode = new_definition `!w:int32. decode w =
       if op then SOME(arm_ZIP2 (QREG' Rd) (QREG' Rn) (QREG' Rm) esize datasize)
       else SOME(arm_ZIP1 (QREG' Rd) (QREG' Rn) (QREG' Rm) esize datasize)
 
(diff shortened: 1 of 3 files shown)


aqjune-aws (approved):
I did the following checks:
- The formalized semantics is consistent with explanation of the doc
- I ran the cosimulator for 30 mins with TBL left only, and it did not find any discrepancy
- TBL is a constant-time instruction! ([this doc](https://developer.arm.com/documentation/ddi0596/2021-03/SIMD-FP-Instructions/TBL--Table-vector-Lookup-) says "The execution time of this instruction is independent of:
 .. The values of the data supplied in any of its registers.")

PR #220, opened 2025-04-07
Extend model of SLI to 64-bit variant

hanno-becker:
Previously, the Arm64 model for the SLI (Vector) instruction only supported the 128-bit variant.

This commit extends the Arm64 model with support for the 64-bit variant of the SLI (Vector) instruction.


hanno-becker:
@jargh @aqjune-aws  Could one of you take a look?

aqjune-aws:
Hi Hanno, do you want to update the simulator as well? It will have SLI with q=1 case enabled only.

hanno-becker:
@aqjune-aws Good catch, thank you. I changed the simulator to test q=0 as well.

aqjune-aws (approved):
LGTM, thanks!

PR #221, opened 2025-04-07
DRAFT: Add ST3 model for datasize=64, esize=8

hanno-becker:
This commit extends the Arm architecture model by support for a specific version of the ST3 instruction: Only datasize=64 and esize=8 are supported.

The matching mnemonic is

```
ST3 {out0.8b, out1.8b, out2.8b}, [dst] { , #24 }
```

### DIFF
--- a/arm/proofs/decode.ml
+++ b/arm/proofs/decode.ml
@@ -94,6 +94,10 @@ let arm_ldstp_2q = new_definition `arm_ldstp_2q ld Rt =
 let arm_ldst2 = new_definition `arm_ldst2 ld Rt =
   let Rtt:(5 word) = word ((val Rt + 1) MOD 32) in
   (if ld then arm_LD2 else arm_ST2) (QREG' Rt) (QREG' Rtt)`;;
+let arm_st3 = new_definition `arm_st3 Rt =
+  let Rtt:(5 word) = word ((val Rt + 1) MOD 32) in
+  let Rttt:(5 word) = word ((val Rt + 2) MOD 32) in
+  arm_ST3 (QREG' Rt) (QREG' Rtt) (QREG' Rttt)`;;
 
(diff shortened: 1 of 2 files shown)


hanno-becker:
@jargh Before adding further variants: Is there a smarter way to model this rather than spelling out the zipping lane by lane?

rod-chapman:
Which specific algorithm/function needs this variant? In the mlkem-native sources for poly_tobytes() I see
```
st3 {out0.8b, out1.8b, out2.8b}, [dst], #24
```
is that variant also on our shopping list?

hanno-becker:
@rod-chapman Yes, that variant is covered here.

rod-chapman:
Lovely... that will be a good one to knock off...

hanno-becker:
@jargh Is there a better way for this?

jargh:
I think the basic approach is right, but note that `word_join`, for better or worse, is defined in a big-endian argument order `word_join high low`, so I think the instances may need swapping. I agree with your implicit comment that it's a shame to have all the definitions like this and the existing `word_interleave2`, `word_interleave4` etc. One idea would be a generic definition like this with a list of arguments and some parameters inferred from the length:
```
let word_interleave = define
 `(word_interleave:num->(N word)list->M word) n inps =
    let l = LENGTH inps in
    word_of_bits { i | let q = i DIV n and r = i MOD n in
                       bit (n * q DIV l + r) (EL (q MOD l) inps)}`;;
```
then one could use `word_interleave 8 [x;y;z]` with appropriate type constraints. That looks more compact and easily validated, but it won't by default get nicely executed unless it is supported with extra conversions (it could even be put in the HOL Light word library and supported fully). But we could simply prove all the existing special cases as theorems and still use those forms when simulating.

Issue #167, opened 2024-12-02
`const` quantifiers in s2n-bignum header file


aqjune-aws:
Currently the s2n-bignum header files do not eagerly attach the `const` quantifier to buffer arguments, but this causes users to introduce const casts.
The customized version of AWS-LC has const already attached: https://github.com/aws/aws-lc/blob/main/third_party/s2n-bignum/include/s2n-bignum_aws-lc.h

Probably we can automatize generation of the function signatures from specifications.

PR #222, opened 2025-04-09
Const-correct header files

torben-hansen:
*Description of changes:*

To enable an improved import process for AWS-LC in https://github.com/aws/aws-lc/pull/2324, we must first const-correct declarations in s2n-bignum. This PR does that.

It's implemented by:
* Identifying parameters that needs a const-correct.
* Running the script below.
* Copy output to `s2n-bignum.h`

The patterns I found that needs const-correct are:
```
* type  x[
* type y[
* type point[
* type  *x
* type  *y
* type  p1[
* type  p2[
* type *table
* type scalar[
* type bscalar[
* type p[
* type c[
* type pp[
* type *m
* type *a
* type *xs
```

The script below implements the translation
```
import re

def process_header_file(input_file, output_file):
    # Read the input file
    with open(input_file, 'r') as f:
        lines = f.readlines()

    # Process each line
    modified_lines = []
    for line in lines:
        # Skip comment lines
        if line.strip().startswith('//') or line.strip().startswith('/*'):
            modified_lines.append(line)
            continue

        # Patterns for parameters we want to modify
        patterns = [
            (r'([a-zA-Z0-9_]+)(\s+)x\[', r'const \1\2x['),
            (r'([a-zA-Z0-9_]+)(\s+)y\[', r'const \1\2y['),
            (r'([a-zA-Z0-9_]+)(\s+)point\[', r'const \1\2point['),
            (r'([a-zA-Z0-9_]+)(\s+)\*x(?=[,\)])', r'const \1\2*x'),
            (r'([a-zA-Z0-9_]+)(\s+)\*y(?=[,\)])', r'const \1\2*y'),
            (r'([a-zA-Z0-9_]+)(\s+)p1\[', r'const \1\2p1['),
            (r'([a-zA-Z0-9_]+)(\s+)p2\[', r'const \1\2p2['),
            (r'([a-zA-Z0-9_]+)(\s+)\*table(?=[,\)])', r'const \1\2*table'),
            (r'([a-zA-Z0-9_]+)(\s+)scalar\[', r'const \1\2scalar['),
            (r'([a-zA-Z0-9_]+)(\s+)bscalar\[', r'const \1\2bscalar['),
            (r'([a-zA-Z0-9_]+)(\s+)p\[', r'const \1\2p['),
            (r'([a-zA-Z0-9_]+)(\s+)c\[', r'const \1\2c['),
            (r'([a-zA-Z0-9_]+)(\s+)pp\[', r'const \1\2pp['),
            (r'([a-zA-Z0-9_]+)(\s+)\*m(?=[,\)])', r'const \1\2*m'),
            (r'([a-zA-Z0-9_]+)(\s+)\*a(?=[,\)])', r'const \1\2*a'),
            (r'([a-zA-Z0-9_]+)(\s+)\*b(?=[,\)])', r'const \1\2*b'),
            (r'([a-zA-Z0-9_]+)(\s+)\*xs(?=[,\)])', r'const \1\2*xs'),
        ]

        # Apply each pattern to the line
        modified_line = line
        for pattern, replacement in patterns:
            modified_line = re.sub(pattern, replacement, modified_line)
        
        modified_lines.append(modified_line)

    # Write the modified content
    with open(output_file, 'w') as f:
        f.writelines(modified_lines)

# Usage
input_file = 'include/s2n-bignum.h'
output_file = 's2n-bignum_const_corrected.h'
process_header_file(input_file, output_file)
```

### DIFF
--- a/include/s2n-bignum.h
+++ b/include/s2n-bignum.h
@@ -520,7 +520,7 @@ extern void bignum_moddouble (uint64_t k, uint64_t *z, const uint64_t *x, const 
 
 // Modular exponentiation for arbitrary odd modulus, z := (a^p) mod m
 // Inputs a[k], p[k], m[k]; output z[k], temporary buffer t[>=3*k]
-extern void bignum_modexp(uint64_t k,uint64_t *z, const uint64_t *a,uint64_t *p,const uint64_t *m,uint64_t *t);
+extern void bignum_modexp(uint64_t k,uint64_t *z, const uint64_t *a,const uint64_t *p,const uint64_t *m,uint64_t *t);
 
 // Compute "modification" constant z := 2^{64k} mod m
 // Input m[k]; output z[k]; temporary buffer t[>=k]
--- a/include/s2n-bignum.h
+++ b/include/s2n-bignum.h
@@ -668,7 +668,7 @@ extern void bignum_mux_6 (uint64_t p, uint64_t z[S2N_BIGNUM_STATIC 6],const uint
 
 // Select element from 16-element table, z := xs[k*i]
 // Inputs xs[16*k], i; output z[k]
-extern void bignum_mux16 (uint64_t k, uint64_t *z, uint64_t *xs, uint64_t i);
+extern void bignum_mux16 (uint64_t k, uint64_t *z, const uint64_t *xs, uint64_t i);
 
 // Negate modulo p_25519, z := (-x) mod p_25519, assuming x reduced
 // Input x[4]; output z[4]
@@ -941,8 +941,8 @@ extern void curve25519_x25519base_byte_alt(uint8_t res[S2N_BIGNUM_STATIC 32],con
 
 // Decode compressed 256-bit form of edwards25519 point
 // Input c[32] (bytes); output function return and z[8]
-extern uint64_t edwards25519_decode(uint64_t z[S2N_BIGNUM_STATIC 8], const const uint8_t c[S2N_BIGNUM_STATIC 32]);
-extern uint64_t edwards25519_decode_alt(uint64_t z[S2N_BIGNUM_STATIC 8], const const uint8_t c[S2N_BIGNUM_STATIC 32]);
+extern uint64_t edwards25519_decode(uint64_t z[S2N_BIGNUM_STATIC 8], const uint8_t c[S2N_BIGNUM_STATIC 32]);
+extern uint64_t edwards25519_decode_alt(uint64_t z[S2N_BIGNUM_STATIC 8], const uint8_t c[S2N_BIGNUM_STATIC 32]);
 
 // Encode edwards25519 point into compressed form as 256-bit number
 // Input p[8]; output z[32] (bytes)
--- a/include/s2n-bignum.h
+++ b/include/s2n-bignum.h
@@ -32,133 +32,133 @@
 
 // Add, z := x + y
 // Inputs x[m], y[n]; outputs function return (carry-out) and z[p]
-extern uint64_t bignum_add (uint64_t p, uint64_t *z, uint64_t m, uint64_t *x, uint64_t n, uint64_t *y);
+extern uint64_t bignum_add (uint64_t p, uint64_t *z, uint64_t m, const uint64_t *x, uint64_t n, const uint64_t *y);
 
 // Add modulo p_25519, z := (x + y) mod p_25519, assuming x and y reduced
 // Inputs x[4], y[4]; output z[4]
-extern void bignum_add_p25519 (uint64_t z[S2N_BIGNUM_STATIC 4], uint64_t x[S2N_BIGNUM_STATIC 4], uint64_t y[S2N_BIGNUM_STATIC 4]);
+extern void bignum_add_p25519 (uint64_t z[S2N_BIGNUM_STATIC 4], const uint64_t x[S2N_BIGNUM_STATIC 4], const uint64_t y[S2N_BIGNUM_STATIC 4]);
 
 // Add modulo p_256, z := (x + y) mod p_256, assuming x and y reduced
 // Inputs x[4], y[4]; output z[4]
-extern void bignum_add_p256 (uint64_t z[S2N_BIGNUM_STATIC 4], uint64_t x[S2N_BIGNUM_STATIC 4], uint64_t y[S2N_BIGNUM_STATIC 4]);
+extern void bignum_add_p256 (uint64_t z[S2N_BIGNUM_STATIC 4], const uint64_t x[S2N_BIGNUM_STATIC 4], const uint64_t y[S2N_BIGNUM_STATIC 4]);
 
 // Add modulo p_256k1, z := (x + y) mod p_256k1, assuming x and y reduced
 // Inputs x[4], y[4]; output z[4]
-extern void bignum_add_p256k1 (uint64_t z[S2N_BIGNUM_STATIC 4], uint64_t x[S2N_BIGNUM_STATIC 4], uint64_t y[S2N_BIGNUM_STATIC 4]);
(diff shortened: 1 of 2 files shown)


pennyannn (commented):
I like this change because it improves function contracts for the exported assembly. Should the couple `word_xxx` functions at the end of the file be updated too?

torben-hansen:
> Should the couple `word_xxx` functions at the end of the file be updated too?

C is pass-by-value. So, there is no value from a compiler optimisation point of view; from the function view, the argument is just merely a local variable, which the compiler can derive whether changes or not.

For that reason, I haven't added a `const` modifier to any non-pointer parameters.

pennyannn:
> > Should the couple `word_xxx` functions at the end of the file be updated too?
> 
> 
> 
> C is pass-by-value. So, there is no value from a compiler optimisation point of view; from the function view, the argument is just merely a local variable, which the compiler can derive whether changes or not.
> 
> 
> 
> For that reason, I haven't added a `const` modifier to any non-pointer parameters.
> 
> 

Sorry, I didn't realize that the word functions use non-pointer inputs. This looks good to me.

pennyannn:
Should this file be updated as well? https://github.com/awslabs/s2n-bignum/blob/main/include/s2n-bignum-c89.h

torben-hansen:
> Should this file be updated as well? https://github.com/awslabs/s2n-bignum/blob/main/include/s2n-bignum-c89.h

We aren't using that file right now.

jargh:
Even so, some customers other than AWS-LC might want this version of the headers; it was originally motivated by being able to replace some arithmetic in https://github.com/bitcoin-core/secp256k1, though this was just an experiment. I think the `const` qualifier was there in C89 so it makes sense to update it too.

However, possibly another change that would subsume that would be to autogenerate `s2n-bignum-c89.h` from `s2n-bignum.h`, which should be fairly trivial and would save editing in the future. So maybe we'll just do that and inherit your `const` changes automatically :-)

torben-hansen:
> Even so, some customers other than AWS-LC might want this version of the headers; it was originally motivated by being able to replace some arithmetic in https://github.com/bitcoin-core/secp256k1, though this was just an experiment. I think the `const` qualifier was there in C89 so it makes sense to update it too.
> 
> However, possibly another change that would subsume that would be to autogenerate `s2n-bignum-c89.h` from `s2n-bignum.h`, which should be fairly trivial and would save editing in the future. So maybe we'll just do that and inherit your `const` changes automatically :-)

Sure. It can be done in another PR.

jargh:
Indeed, let's save the generation of the C89 version for another PR. I have cross-checked this one somewhat carefully including some scripted tests, and only noticed two possible issues:
* In `bignum_mux16` the `xs` argument should be `const`
* In `edwards25519_decode` and `edwards25519_decode_alt`, there is a `const const`, which I guess is not wrong but a bit redundant/confusing.

aqjune-aws:
This is nice..! Is it legit to declare #167 as fixed after this patch as well?

jargh:
I think so, indeed. One little thing still: although I am convinced this is right, it would be nice to update t`ests/test.c` accordingly to fix various compiler warnings that now arise. On gcc they are just annoying, but on clang on my Mac they actually make the compilation fail:
`$ make
gcc -Wall -std=gnu99 -o test test.c -L../arm -ls2nbignum
test.c:3180:13: error: incompatible function pointer types passing 'void (uint64_t *, const uint64_t *, uint64_t, uint64_t, uint64_t)' (aka 'void (unsigned long long *, const unsigned long long *, unsigned long long, unsigned long long, unsigned long long)') to parameter of type 'void (*)(uint64_t *, uint64_t *, uint64_t, uint64_t, uint64_t)' (aka 'void (*)(unsigned long long *, unsigned long long *, unsigned long long, unsigned long long, unsigned long long)') [-Wincompatible-function-pointer-types]
 3180 |       0, 0, bignum_copy_row_from_table);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~
test.c:3131:12: note: passing argument to parameter 'f' here
 3131 |     void (*f)(uint64_t*, uint64_t*, uint64_t, uint64_t, uint64_t))
      |            ^
`

torben-hansen:
> I think so, indeed. One little thing still: although I am convinced this is right, it would be nice to update `tests/test.c` accordingly to fix various compiler warnings that now arise. On gcc they are just annoying, but on clang on my Mac they actually make the compilation fail:
> 
> ```
> $ make
> gcc -Wall -std=gnu99 -o test test.c -L../arm -ls2nbignum
> test.c:3180:13: error: incompatible function pointer types passing 'void (uint64_t *, const uint64_t *, uint64_t, uint64_t, uint64_t)' (aka 'void (unsigned long long *, const unsigned long long *, unsigned long long, unsigned long long, unsigned long long)') to parameter of type 'void (*)(uint64_t *, uint64_t *, uint64_t, uint64_t, uint64_t)' (aka 'void (*)(unsigned long long *, unsigned long long *, unsigned long long, unsigned long long, unsigned long long)') [-Wincompatible-function-pointer-types]
>  3180 |       0, 0, bignum_copy_row_from_table);
>       |             ^~~~~~~~~~~~~~~~~~~~~~~~~~
> test.c:3131:12: note: passing argument to parameter 'f' here
>  3131 |     void (*f)(uint64_t*, uint64_t*, uint64_t, uint64_t, uint64_t))
>       |            ^
> ```

That is now fixed.

(should add a clang dimension to CI)

torben-hansen:
> This is nice..! Is it legit to declare #167 as fixed after this patch as well?

I'll leave that to you. Having "Probably we can automatize generation of the function signatures from specifications." would have saved a lot of work, though.

include/s2n-bignum.h
```diff
@@ -175,769 +175,769 @@ extern void bignum_copy_row_from_table (uint64_t *z, uint64_t *table, uint64_t h
...
+extern uint64_t bignum_digit (uint64_t k, const uint64_t *x, uint64_t n);
 
 // Return size of bignum in digits (64-bit word)
 // Input x[k]; output function return
-extern uint64_t bignum_digitsize (uint64_t k, uint64_t *x);
+extern uint64_t bignum_digitsize (uint64_t k, const uint64_t *x);
 
 // Divide bignum by 10: z' := z div 10, returning remainder z mod 10
 // Inputs z[k]; outputs function return (remainder) and z[k]
 extern uint64_t bignum_divmod10 (uint64_t k, uint64_t *z);
```
aqjune-aws, line 263: Should z be constant here as well?

include/s2n-bignum.h
```diff
@@ -175,769 +175,769 @@ extern void bignum_copy_row_from_table (uint64_t *z, uint64_t *table, uint64_t h
...
 
 // Double modulo m, z := (2 * x) mod m, assuming x reduced
 // Inputs x[k], m[k]; output z[k]
-extern void bignum_moddouble (uint64_t k, uint64_t *z, uint64_t *x, uint64_t *m);
+extern void bignum_moddouble (uint64_t k, uint64_t *z, const uint64_t *x, const uint64_t *m);
 
 // Modular exponentiation for arbitrary odd modulus, z := (a^p) mod m
 // Inputs a[k], p[k], m[k]; output z[k], temporary buffer t[>=3*k]
-extern void bignum_modexp(uint64_t k,uint64_t *z, uint64_t *a,uint64_t *p,uint64_t *m,uint64_t *t);
+extern void bignum_modexp(uint64_t k,uint64_t *z, const uint64_t *a,uint64_t *p,const uint64_t *m,uint64_t *t);
```
aqjune-aws, line 523: Should p be constant here as well?

include/s2n-bignum.h
aqjune-aws: Sorry, z is also an output of this function. :/

jargh:
All looks very good to me now, and I'll merge. Thanks!

PR #223, opened 2025-04-10
Tutorial for instruction modeling

pennyannn:
*Description of changes:*

This PR adds tutorials for instruction modeling on Arm and x86_64.


aqjune-aws:
Hi Yan, I was contemplating about how to add the DOIT list check for x86. (discussed in #270)
Since s2n-bignum already has the 'allowed_asm' file, it seemed having another list of instructions to check was redundant.
Instead, what do you think about adding to x86/INSTRUCTION.md that one must check whether the new instruction appears from the [DOIT list](https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/resources/data-operand-independent-timing-instructions.html)?
As an official guideline document for adding instructions, this seems to be a really good place to add it I think.

pennyannn:
> Hi Yan, I was contemplating about how to add the DOIT list check for x86. (discussed in #270) Since s2n-bignum already has the 'allowed_asm' file, it seemed having another list of instructions to check was redundant. Instead, what do you think about adding to x86/INSTRUCTION.md that one must check whether the new instruction appears from the [DOIT list](https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/resources/data-operand-independent-timing-instructions.html)? As an official guideline document for adding instructions, this seems to be a really good place to add it I think.

Hi June, I added a section in x86/INSTRUCTION.md to talk about this.

aqjune-aws (approved):

PR #224, opened 2025-04-16
Pre-reduce nested subword expressions in SIMD semantics on ARM

jargh:
This just applies WORD_SIMPLE_SUBWORD_CONV to the semantic definitions of most of the ARM NEON instructions, since the way they are expressed results in a lot of redundancy. It seems clearly better to remove this once and for all rather than per use. Only two proofs, both of which are really minor variants of the same one (proving the Lenngren-style X25519 code) needed to be modified. Otherwise this just results in small but worthwile speedups, particularly significant for instruction cosimulation (which seems about 5% faster).

*Issue #, if available:*

*Description of changes:*


aqjune-aws (approved):
I checked the improvement of the cosimulator speed from Codebuild log. :) # reg instructions improved from 1000~1200 to 1100~1300.

PR #226, opened 2025-04-19
Extend Mach-O reader to read rodata and relocation entries

aqjune-aws:
This patch extends the Mach-O reader of s2n-bignum to extract the read-only data (the `__const` section) and relocation entries for `ADRP`/`ADD` as well.

It extends the `load_macho` function in common/elf to return the new information and makes its function signature same as `load_elf`. The `arm/tutorial/rodata.{S,ml}` tutorial was updated. It should work on Mac. The new curve25519_x25519base.ml that John wrote also passed with his curve25519_x25519base.S was cross-compiled to Mach-O as well!

Materials that were useful to me were:
- OS X ABI Mach-O File Format Reference: https://github.com/aidansteele/osx-abi-macho-file-format-reference
- Relocation entry types from MachO.h of LLVM: https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/BinaryFormat/MachO.h#L451

I also found that https://github.com/horsicq/XMachOViewer was useful for its visualization.

One interesting thing from Mach-O's symbol table was that it has 'virtual' symbols whose names start with 'ltmp'. I found that other MachO dump tools ignore these symbols (one of those was https://github.com/microsoft/llvm-mctoll/blob/master/MachODump.cpp#L228 and another one was from another open source tool), so I chose to mimic their choice :)

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/arm/tutorial/rodata.S
+++ b/arm/tutorial/rodata.S
@@ -16,12 +16,15 @@
   }
 */
 
-.text
+#if defined(__linux__) && defined(__ELF__)
 .section  .rodata
   .global  x
-  .align  3
   .type  x, %object
   .size  x, 40
+#elif defined(__APPLE__)
+.const_data
+#endif
+  .align  3
 x:
   .word  2
   .word  4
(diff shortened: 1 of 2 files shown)


jargh (approved):
This is great, thanks! I've tested both the tutorial example and an updated version of `curve25519_x25519base` using `rodata` and it now works on my Mac without problems.

PR #228, opened 2025-04-25
Support lockstep simulation of XMM_SSE registers, add coda param to mk_equiv_statement, and minor refactorings

aqjune-aws:
1. Support lockstep simulation of XMM_SSE registers, add READ_YMM_SSE_EQUIV to `extra_word_CONV`

- This patch extends X86_LOCKSTEP_TAC to support SSE registers.
Previously their lockstep simulation could not be achieved for XMM_SSE
when there was no precondition on the equivalence of upper halfs of the
YMM_SSE registers, because vec operations on XMM_SSE preserves the upper halfs.
This patch relaxes X86_LOCKSTEP_TAC so that if the two output registers of
the lockstep-ping instructions are having `word_join x y` and
`word_join x' y'` as their values, and `y = y'` could be proven true,
then successfully abbreviate `y` and `y'` to the same temporary
variables.

- Also, this commit has the following updates:

  * Add `READ_YMM_SSE_EQUIV` to `extra_word_CONV`. This theorem rewrites
    `read YMM{n}_SSE ..` to `read YMM{n} ..`. This is a useful
    canonicalization and in fact was already existing in `X86_STEP_TAC`.\
    This patch removes the explicit `REWRITE_TAC READ_YMM_SSE_EQUIV` in
    `X86_STEP_TAC` because the `REWRITE_TAC` could rewrite post conditions
    as well.
  * Improves the error message printing in `X86_CONV` and `ARM_CONV` when
    there is no `{aligned_}bytes_loaded` or `read {PC|RIP}` in
    assumptions.

2.  Combine mk_equiv_statement in x86 and arm 

3. Add an extra arg for data in text to mk_equiv_statement 

- This third patch is mainly written by Yan Peng.

- This adds an extra parameter to `mk_equiv_statement` that optionally
describes the constant data. The constant data is placed just after the
machine code in the text section. The output of
`define_coda_literal_from_elf` can be forwarded to these extra params.

--

### DIFF
--- a/arm/curve25519/bignum_add_p25519.S
+++ b/arm/curve25519/bignum_add_p25519.S
@@ -5,8 +5,8 @@
 // Add modulo p_25519, z := (x + y) mod p_25519, assuming x and y reduced
 // Inputs x[4], y[4]; output z[4]
 //
-//    extern void bignum_add_p25519
-//     (uint64_t z[static 4], uint64_t x[static 4], uint64_t y[static 4]);
+//    extern void bignum_add_p25519(uint64_t z[static 4], const uint64_t x[static 4],
+//                                  const uint64_t y[static 4]);
 //
 // Standard ARM ABI: X0 = z, X1 = x, X2 = y
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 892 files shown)


pennyannn:
Thank you June for this thoughtful patch! Can't wait to try it out!

pennyannn:
I have used features provided in the PR in my proof and it worked well. Maybe we should merge in the changes before more features are requested and the PR explodes.:-)

pennyannn (approved):

PR #230, opened 2025-04-30
Make s2n-bignum _CET_ENDBR play nice with downstream

torben-hansen:
*Description of changes:*

Downstream can define `_CET_ENDBR` if `cet.h` is not present. And sometimes, in downstream, s2n-bignum source code is not compiled in isolated s2n-bignum-only compilation units. The result of those two facts is redefinitions of `_CET_ENDBR`.

Prevent this by guarding the definition of `_CET_ENDBR` with a is-it-already-defined-? test.

Also, make sure that an inconsistent build configuration cannot proceed i.e. `NO_IBT` is defined but `_CET_ENDBR` is also defined... I believe this should be a fatal build error, so I made it that.

### DIFF
--- a/include/_internal_s2n_bignum.h
+++ b/include/_internal_s2n_bignum.h
@@ -33,9 +33,12 @@
 // slowdown from executing one more instruction.
 
 #if NO_IBT
-#define _CET_ENDBR
+#   if defined(_CET_ENDBR)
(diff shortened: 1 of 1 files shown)


jargh (approved):
This all makes good sense and passes my tests, thanks! I will merge it.

PR #231, opened 2025-05-13
Add AVX2 instructions vpsubw, vpaddw, vpmullw and vpmulhw

pennyannn:
*Description of changes:*

This PR adds AVX2 instructions vpsubw, vpaddw, vpmullw, and vpmulhw. In addition, an issue is fixed in the `read_VEX` function for reading 2-byte VEX.

### DIFF
--- a/x86/proofs/decode.ml
+++ b/x86/proofs/decode.ml
@@ -258,7 +258,8 @@ let read_VEXP = new_definition `read_VEXP (p:2 word) =
 let read_VEX = define
  `(!l. read_VEX T l =
    read_byte l >>= \(b,l). bitmatch b with [r:1; v:4; L; p:2] ->
-   SOME((SOME(word_zx (word_not r)), VEXM_0F, word_not v, L, (F, Rep0, SG0)), l)) /\
+   SOME((SOME(word_shl (word_zx (word_not r)) 2), 
+     VEXM_0F, word_not v, L, (F, Rep0, SG0)), l)) /\
   (!l. read_VEX F l =
    read_byte l >>= \(b,l). bitmatch b with [rxb:3; m:5] ->
    read_byte l >>= \(b,l). bitmatch b with [w; v:4; L; p:2] ->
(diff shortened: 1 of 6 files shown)


x86/proofs/decode.ml
```diff
@@ -258,7 +258,8 @@ let read_VEXP = new_definition `read_VEXP (p:2 word) =
 let read_VEX = define
  `(!l. read_VEX T l =
    read_byte l >>= \(b,l). bitmatch b with [r:1; v:4; L; p:2] ->
-   SOME((SOME(word_zx (word_not r)), VEXM_0F, word_not v, L, (F, Rep0, SG0)), l)) /\
+   SOME((SOME(word_shl (word_zx (word_not r)) 2),
```
pennyannn, line 261: Previously the REX.R bit is not set in the 2th location of the half-byte. The added shift fix that.

x86/proofs/decode.ml
```diff
@@ -1624,6 +1641,7 @@ let READ_SIB_CONV,READ_MODRM_CONV,READ_VEX_CONV,DECODE_CONV =
     evaluate (rhs (concl th)) (F o TRANS th)
   | Comb(Const("@",_),_) -> failwith "ARB"
   | Const("ARB",_) -> failwith "ARB"
+  | Comb(Comb((Const("word_shl",_) as f),a),b) -> eval_binary f a b F WORD_SHL_CONV
```
pennyannn, line 1644: The evaluator needs to understand `word_shl` and `NUMERAL` for the REX.R code in `read_VEX`.

jargh (approved):
This is all great! Thanks also for catching the `read_VEX` issue and to June for adding the new `weakly_valid_component` theorems.

One little suggestion would be to add some `xmm` cases to the `iclasses` list of extra instructions in `x86/proofs/simulator.ml` as well as the `ymm` ones you've already added, to make sure we are covering all paths in the cosimulation even before we make use of these instructions in real code. But I've done some additional checks myself and have no concerns about the xmm cases being right, so I'll merge this now so that we can start making use of it :-)

Issue #209, opened 2025-03-18
Speedup ARM_ADD_RETURN_STACK_TAC


hanno-becker:
For some proofs in mlkem-native, specifically those introduced in https://github.com/pq-code-package/mlkem-native/pull/883, I notice that ARM_ADD_RETURN_STACK_TAC takes far longer than the actual correctness proof for the body of the function. The worst one is `mlkem_poly_basemul_acc_montgomery_cached_k4`, which takes 2h in CI, mostly in  `ARM_ADD_RETURN_STACK_TAC` outside of symbolic execution.

It would be nice to analyze where time is lost here, and improve it.

hanno-becker:
One step that seems surprisingly slow is `ENSURES_EXISTING_PRESERVED_TAC`, and probably also `ENSURES_PRESERVED_DREG_TAC`. Within `ENSURES_EXISTING_PRESERVED_TAC`, one faces a goal of the shape
```
ASSIGNS( c0 ,, c1 ,, ... ,, cN) s s' ==> f s = f s'
```
where `f` is a component orthogonal to all `c_i`. This is approached using
```
REWRITE_TAC[ASSIGNS_SEQ] 
```
which applies `ASSIGNS_SEQ: |- (ASSIGNS c ,, r) = \s s'. ?y. r (write c y s) s'` to generate `N` nested existentials for the modifications of `s` into `s'`. Somewhat surprisingly, this takes noticable time.

I am not sufficiently familiar with HOL-Light yet to proceed on my own, but it would seem that an alternative approach could be to prove something along the lines of
```
let ASSIGNS_ORTHOGONAL = `
        (!(s:armstate) x. (f : armstate -> 'a) (write c s x) = f s) ==>
        (!(s:armstate) s'. r s s' ==> f s' = f s) ==>
        (ASSIGNS c ,, r) s s' ==> f s' = f s
        `;;
```
and use that to generate N separate (non-nested) orthogonality goals for `f` and the `c_i`.

hanno-becker:
The following helps a little. With 
```
let SEQ_ORTHOGONAL = prove(`! s s' r0 r1 f.
        ((!(x:armstate) x'. r0 x x' ==> (f : armstate -> 'a) x' = f x) /\
        (!(x:armstate) x'. r1 x x' ==> f x' = f x)) ==>
        ( r0 ,, r1 ) s s' ==> f s' = f s
[...]

PR #240, opened 2025-06-13
Fix slow speeds of tactics using REWRITE_TAC[ASSIGNS_SEQ]

aqjune-aws:
This patch fixes slow tactics that use `REWRITE_TAC[ASSIGNS_SEQ]`. `ASSIGNS_SEQ` introduces lambda functions to the expression, and then `REWRITE_TAC` applies the beta rule among `basic_rewrites()`. It seems this alternating application of two different rules cause significant slowdown, especially when `ASSIGNS_SEQ` introduces many nested expressions.
Experimental result is here:
https://gist.github.com/aqjune-aws/7237f7f082fcc6010e5bbe6cc614acff

Using `PURE_REWRITE_TAC[ASSIGNS_SEQ] THEN CONV_TAC (TOP_DEPTH_CONV BETA_CONV)` makes it faster. For the specific example that I attached in Gist, this reduces running time from 7.6 sec to 0.03.

*Issue #, if available:*

#209 

*Description of changes:*


jargh (approved):
This is very nice, thank you! I'm actually quite surprised that it makes such a difference :-)

PR #241, opened 2025-06-14
Add REV to ARM model

amanda-zx:
*Issue #, if available:* N/A

*Description of changes:*
Added the REV instruction to the ARM model, including its decoding and its co-simulation against hardware.


aqjune-aws:
There are other pending pull requests before this PR was made, but I believe they are either (1) x86 parts, or (2) updates to Arm that is not related to this PR, so I hope that this does not interfere with the PRs... I will carefully merge this.

PR #248, opened 2025-06-26
Add AVX2 VPSRAD instruction

jakemas:
*Issue #, if available:*
N/A

*Description of changes:*
This change models the AVX2 VPSRAD instruction. This is based on the reference documentation which shows that `VPSRAD` uses opcode `0x72` with the `/4` encoding (which corresponds to the `0b100` bit pattern).


x86/proofs/simulator.ml
```diff
@@ -318,6 +318,9 @@ let iclasses = iclasses @
  [0xc4; 0xc1; 0x6d; 0x71; 0xd0; 0x28;]; (* VPSRLW (%_% ymm2) (%_% ymm8) (Imm8 (word 40)) *)
  [0xc5; 0xe5; 0x71; 0xd7; 0x37;]; (* VPSRLW (%_% ymm3) (%_% ymm7) (Imm8 (word 55)) *)
  [0xc4; 0xc1; 0x05; 0x71; 0xd7; 0xff]; (* VPSRLW (%_% ymm15) (%_% ymm15) (Imm8 (word 255)) *)
+ [0xc5; 0xfd; 0x72; 0xe1; 0x01]; (* VPSRAD ymm0, ymm1, 1 *)
```
pennyannn, line 321: Extend to 8 test cases as is done with other AVX2 instructions. It's better to cover ymm8-ymm15 registers as well because they have a slightly different encoding.

pennyannn (commented):
I've walked through the code for decoding, semantics and tests. Only one minor comment otherwise LGTM.

x86/proofs/simulator.ml
jakemas: Thank you for the review, I have included additional test cases to cover ymm8-ymm15 registers in [459ec72](https://github.com/awslabs/s2n-bignum/pull/248/commits/459ec721b1d67aad08fda2772cea859f6326671e)

PR #255, opened 2025-07-02
Add AVX2 VPBLENDD instruction

jakemas:
*Issue #, if available:*
N/A
*Description of changes:*
This change models the AVX2 Vector Packed Blend Doublewords (`VPBLENDD`) instruction.

Callouts:
I had to add a new section `| VEXM_0F3A ->` within `decode.ml`

### DIFF
--- a/arm/proofs/simulator.ml
+++ b/arm/proofs/simulator.ml
@@ -595,7 +595,7 @@ let run_random_simulation() =
     let decoded, result = run_random_memopsimulation() in
     decoded,result,false;;
 
-let time_limit_sec = 1800.0;;
+let time_limit_sec = 2400.0;;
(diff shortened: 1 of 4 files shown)


jakemas:
Having some issues with this one getting picked up correctly by `decode.ml`. My thoughts are that the new section `VEXM_0F3A ->` may not be implemented correctly.

aqjune-aws (approved):
I confirmed that running the cosimulator with the new instruction patterns ran successfully.

Moved `msimd*` to common/words2.ml, and slightly increased the running times of arm/x86 cosimulators from 30 mins to 45 mins because both of these are having more patterns than before (when it was set to 30min :).

PR #257, opened 2025-07-09
Add AVX2 VPSLLQ instruction

jakemas:
*Issue #, if available:*
N/A
*Description of changes:*
This change models the AVX2 Vector Packed Shift Left Logical Quadwords (`VPSLLQ`) instruction.


jakemas:
```
`exists s'. x86_VPMULDQ YMM0 YMM0 YMM1 (write RIP (word (pc + 5)) s0) s'`
Failed after (user) CPU time of 0.022006: Failure("X86_BASIC_STEP_TAC: Equality between two states is ill-formed. Did you forget to assume an extra condition like pointer alignment?")
Fatal error: exception Failure("X86_BASIC_STEP_TAC: Equality between two states is ill-formed. Did you forget to assume an extra condition like pointer alignment?")
make: *** [Makefile:537: sematest] Error 1
Fatal error: exception Failure("X86_BASIC_STEP_TAC: Equality between two states is ill-formed. Did you forget to assume an extra condition like pointer alignment?")
Simulator 1 failed!
```

jargh:
Ah, it's just the peril of those merges with conflicting lines. (Even though it's verbose, I suppose it would make git merges simpler to use one per line, now that so many people are working on this in parallel.) When adding `x86_VPSLLQ_ALT` to the `X86_OPERATION_CLAUSES` list it _replaced_ `x86_VPMULDQ_ALT`, so the latter isn't getting expanded when encountered. It just needs to be added back alongside the new `x86_VPSLLQ_ALT`

jargh (approved):
All looks very good, thanks! The encoding details are a bit intricate, but I have convinced myself they match the Intel manuals and I'm happy that @pennyannn has also looked at it. I've also exercised the cosimulation tests extensively and they all work perfectly.

PR #262, opened 2025-07-23
Combine tactics for ensures_n with ensures

aqjune-aws:
This patch refactors tactics for `ensures_n` and `ensures` by combining these if possible.
This still leaves a few tactics like `ENSURES_N_SEQUENCE_TAC` unmerged which has to take additional argument(s) for number of steps, unlike `ENSURES_SEQUENCE_TAC`.

The refactoring is done as follows.

(1) `common/relational_n.ml` is now loaded from `common/relational.ml`, rather than separately from `{arm,x86}/proofs/base.ml`, because tactics of `common/relational.ml` now must be aware of the existence of `ensures_n`.
(2) Defines two helper functions `ENSURES_OR_ENSURES_N_TAC tac1 tac2`, which chooses the right tactic according to the conclusion in goal, as well as `ensures_or_ensures_n (t:term) x y` that has a similar function
(3) Rewrites tactics for `ensures/eventually` by properly choosing the right theorem/tactic using these two helper functions.

*Issue #, if available:*

*Description of changes:*


jargh (approved):
This is a nice rationalization, thank you!

PR #266, opened 2025-08-07
Add collect-signatures.py that collects input/output params from C signatures

aqjune-aws:
This patch adds tools/collect-signatures.py that
(1) collects input/output params from C function signatures and its comments in s2n-bignum.h (2) stores the information at {arm,x86}/proofs/subroutine_signatures.ml .

The python script also sanity-checks whether there is any discrepancy between the comments in assembly and s2n-bignum.h.
As a result, this commit also contains edits to a few assembly comments as well as edits to comments in s2n-bignum.

This is the first step for upstreaming the effort of semi-automatically proving constant-time and memory safety properties in s2n-bignum.

Also, Makefile is mildly edited to add (a subset of) dependencies from the proofs to the foundational .ml files as well as the new subroutine_signatures.ml .

*Issue #, if available:*

*Description of changes:*


jargh (approved):
Thank you June! I've finally taken a proper close look at this, and it all works really nicely. It's also nice that you caught so many other documentation/comment issues as a side-effect.

arm/proofs/subroutine_signatures.ml
nebeid, line 1: I think it would be good to put as a comment at the top of this file and the x86 one that they are automatically generated from `subroutine_signatures.ml`.

PR #268, opened 2025-08-19
X86 Keccak Proofs

manastasova:
This PR adds an x86 assembly implementation of the Keccak-f1600 function with:

- x86_64 assembly code
- Formal correctness proof of the main function
- Stack- and IBT-enhanced proofs
    
This PR does _not_ add Windows calling convention support

Tests and benchmarks have been updated to validate the implementation and measure performance.


jargh:
Many thanks for all the updates - it now looks pretty much perfect to me! (It's maybe a bit unfortunate to have the `keccak_spec.ml` file duplicated on the ARM and x86 sides, but let's deal with that, and making some other mlkem-related stuff common later - perhaps it even makes sense to upstream this into HOL Light.) I have only two remaining requests:
* Could you rebase by merging the latest mainline commits just to make sure we're up to date before merging this one?
* Can you double-check whether we're safe asserting the usual copyright, since I know the code is at least partly based on some pre-existing code? if necessary we can add an acknowledgement and/or adjust the license line at the top of the .S file. For example the ARM mirror-image tweaks "MIT-0" to "MIT" at the top and adds a comment about provenance just before the `S2N_BN_SYMBOL`.

Anyway, right now I'll also run a full `make proofs` to double-check, but since I ran the new proof file last night I don't anticipate any problems.

jargh:
The run of`make proofs` was indeed fine, as expected.

PR #271, opened 2025-08-21
Print PC and instruction when {x86,arm}_print_log is set

aqjune-aws:
Also rename decode_log to decode_print_log for consistent naming

*Issue #, if available:*

*Description of changes:*


jargh:
I think this is a great addition! Soon we are going to wonder how we possibly survived without it :-)

My only slight question is whether we should be printing the `pc + n` offsets `n` in decimal or hex. I am somewhat inclined towards hex because it makes it easier to compare with (default) `objdump` output. Any other opinions? Of course, the decimal/hex choice could be controlled by yet another settable flag, but this could be considered as getting too complicated.

PR #272, opened 2025-08-21
Add constant-time and memory safety properties for MLKEM Arm functions

aqjune-aws:
This patch adds constant-time and memory safety properties for MLKEM Arm functions, modulo rejection sampling.

*Issue #, if available:*

*Description of changes:*


jargh:
Very nice, thank you! I like the new elegant statements just quantifying over functions of the public data, as well as the high level of automation in the proofs. I've also confirmed all the proofs run and that the new safety specifications get proved. The slowdown in proof files as a result of the safety part are relatively modest as well. I will merge!

PR #273, opened 2025-08-25
Generalize alignment automation from size 16 to any size

jargh:
The earlier ALIGNED_16_CONV and ALIGNED_16_TAC were specifically targeted at the 16-byte stack alignment conditions arising in the ARM code. But now that we are considering x86 instructions that require 32-byte alignment, this is no longer sufficient. This change generalizes it to handle any specific value in the same way, and to avoid confusion renames from ALIGNED_16_... to ALIGNED_WORD_...

This update also adds a couple more cases to "bytesize" in the x86 model, which is also useful for instruction decoding and analysis when operands have length 256.

*Issue #, if available:*

*Description of changes:*


jargh:
Thank you! Alll the proofs ran fine so I have now merged.

PR #276, opened 2025-08-28
Fallback ARM Keccak option

jargh:
*Issue #, if available:*

*Description of changes:*

This adds yet one more ARM variant of the Keccak permutation, sha3_keccak_f1600_alt2. This is intended as a fallback for (micro)architectures where both the following hold:

* shifted operands are not "free" so the use of lazy rotations in `sha3_keccak_f1600` is an anti-optimization
* the sha3 NEON operations that are used in `sha3_keccak_f1600_alt` are unavailable or slow

This seems to be mainly relatively old uarchs such as A-72 used in the first-generation Graviton. (For Apple M1 lazy rotations are also a pessimization but the use of sha3 NEON instructions is much faster.)

The assembler implementation is simply compiled from C code that is more or less a direct and straightforward translation of the Keccak specification via `gcc -O3 -mcpu=cortex-a72 -ffixed-x18 -c`, with a formal proof that is a simple modification of the existing ones.


jargh:
Thank you both for the review, and also thanks to June for adding the constant-time proof! I had forgotten that, and it is something I should remember going forward, but I'm really happy that we have consistent constant-time proofs across all the Keccak variants.

Although this is not very time-critical for AWS-LC integration, since we've missed the AWS-LC FIPS code freeze for this year, I'm inclined to just merge it today. It has the merit that it is more or less orthogonal to other things and it does also fix one small infrastructure bug (the incorrect switching off of one of the Keccak variants in tests and benchmarks). If there are no objections I'll run some final checks and merge it soon.

PR #279, opened 2025-09-04
ML-DSA poly_reduce proof

jakemas:
This PR adds the first x86_64 proof with AVX2 instruction set; an assembly implementation of the mldsa polynomial reduction (`poly_reduce`) function with:

- x86_64 assembly code with AVX2 instructions --  based on the AVX2 C intrinsics only upstream reference implementation  https://github.com/pq-crystals/dilithium/blob/master/avx2/poly.c#L30-L56
- Formal correctness proof of the main function (`MLDSA_REDUCE_CORRECT`)
- Subroutine call/return proofs (`MLDSA_POLY_REDUCE_NOIBT_SUBROUTINE_CORRECT`, `MLDSA_POLY_REDUCE_SUBROUTINE_CORRECT`)
- IBT-enhanced proofs (Intel Branch Tracking support via `ADD_IBT_RULE`)
- Windows ABI compatibility proofs (`MLDSA_POLY_REDUCE_WINDOWS_SUBROUTINE_CORRECT`)

**Function overview:**

The function `mldsa_poly_reduce(int32_t a[static 256])` performs canonical reduction of 256 signed 32-bit polynomial coefficients modulo 8380417, centered reduction to range [-6283008, 6283008]. The architecture is x86_64 with AVX2 instruction set. 

### **Changes made:**

Core implementation:
- Added `86/mldsa/mldsa_poly_reduce.S` 
- and auto-generated AT&T syntax version `x86_att/mldsa/mldsa_poly_reduce.S`.

Proof file:
- Added `mldsa_poly_reduce.ml`

Make files:
- `x86/Makefile`: Added `mldsa/mldsa_poly_reduce.o` to `BIGNUM_OBJ` list
- `x86_att/Makefile`: Added translation rule for automatic AT\&T syntax generation
- Windows object generation: Integrated with existing `WINOBJ` pattern rules

Header files:
- `include/s2n-bignum.h`: Added function prototype with `S2N_BIGNUM_STATIC` annotation
- `include/s2n-bignum-c89.h`: Added C89-compatible function prototype

Testing:
- Added `mld_reduce32_ref()` reference implementation using ML-DSA centered reduction
- Added `test_mldsa_poly_reduce()`

```
~/sep4/s2n-bignum/tests$ ./test mldsa_poly_reduce
Testing mldsa_poly_reduce with 100 cases
All OK
```

Benchmarks
- Added `call_mldsa_poly_reduce()` wrapper function

```
mldsa_poly_reduce               :     0.1 ns each (var 59.0%, corr -0.02) = 10586319218 ops/sec
ARITHMEAN (  1 tests)    :    0.1 ns
GEOMEAN   (  1 tests)    :    0.1 ns
```


x86/proofs/mldsa_poly_reduce.ml
```diff
@@ -0,0 +1,758 @@
...
+                      ==> ival(read(memory :> bytes32
+                                 (word_add a (word(4 * i)))) s) =
+                          ival(mldsa_barred (x i)))
+             (MAYCHANGE [RSP] ,, MAYCHANGE_REGS_AND_FLAGS_PERMITTED_BY_ABI ,,
+              MAYCHANGE [memory :> bytes(a,1024)])`,
+  MATCH_ACCEPT_TAC(ADD_IBT_RULE MLDSA_POLY_REDUCE_NOIBT_SUBROUTINE_CORRECT));;
+
+(* ------------------------------------------------------------------------- *)
+(* General lemmas to eventually put elsewhere.                               *)
+(* ------------------------------------------------------------------------- *)
```
jakemas, line 510: I can move these to a util/common location if there is a good place you have in mind

jargh: They are pretty x86-specific, so probably a good place would be in `x86/proofs/x86.ml`. Perhaps just after `WINDOWS_MAYCHANGE_REGS_AND_FLAGS_PERMITTED_BY_ABI`, since they are concerned with `MAYCHANGE` and are intended to handle issues around the Windows ABI, even though they are not directly connected. In the fairly near future I'll wrap these up in some proper automation so we won't need to invoke that sort of gruesome proof manually every time, but we might need to try a few more examples first to see what the general case of a Windows wrapper looks like in the case of AVX2 usage.

PR #282, opened 2025-09-06
Fix missing `SUBROUTINE_SAFE` counting, resolve failures when running proofs on Mac OS

aqjune-aws:
*Issue #, if available:*

*Description of changes:*


jargh (approved):
Thank you, this all works very well - I've tried loading object files both with and without rodata in Mach-O and it works smoothly, and a full `make proofs` is successful.

PR #283, opened 2025-09-08
Add definition of uarch event in x86 for describing variable-time instructions

aqjune-aws:
*Issue #, if available:*

*Description of changes:*

1. Factoring out safety-related definitions to common/safety.ml
- Factor out microarch definition and memaccess_inbounds to comon/safety.ml
- rename it from armevent to uarch_event
- From the x86 side, add two non-constant instructions PEXT and POPCNT as constructors of uarch_event (EventX86PEXT and EventX86POPCNT)
2. Updating x86_execute and x86 proofs
- When the operands of an instruction is Memop, add EventStore and EventLoad to the event list. Two helper functions add_store_event and add_load_event are added to do this.
- Add MAYCHANGE [event] to existing functional correctness proofs.
3. Other refactorings
- Remove x86_DIV2 (x86/proofs/instruction.ml update as well) because it is not constant time and not being used (but PEXT and POPCNT which are non-constant-time are added; should I revive this?)
- Removes duplicated VPBROADCASTQ from x86_execute


jargh (approved):
All looks good, thank you! It's nice to have a unified story between ARM and x86.

PR #286, opened 2025-09-09
Add AVX2 VPERMQ instruction

jakemas:
*Issue #, if available:*
MLKEM/MLDSA NTT/iNTT proof

*Description of changes:*


x86/proofs/x86.ml
```diff
@@ -1417,6 +1417,17 @@ let x86_VPSLLQ = new_definition
...
+let x86_VPERMQ = new_definition
+  `x86_VPERMQ dest src imm8 (s:x86state) =
+      let (x:N word) = read src s
+      and od = read imm8 s in
+      if dimindex(:N) = 256 then
+        let res:(256)word = usimd4 (\(od:(2)word).
+            (word_subword:int256->num#num->int64) (word_zx x) ((val od)*64,64)) od in
+        (dest := (word_zx res):N word) s
+      else
+        raise_exception x86_Exception_UD s`;;
```
jakemas, line 1429: Is this the correct way to handle this? 

VPERM2I128 is an AVX2-only instruction that operates exclusively on 256-bit YMM registers. According to the Intel Software Developer's Manual, VPERM2I128 has no 128-bit (XMM) variant - it only exists in the 256-bit form. When the instruction is attempted on 128-bit operands, the processor generates an Undefined Instruction (#UD) exception.

jargh:
I think the approach of raising an exception is fine and perhaps conforms best to the intent of the manual. It would also be possible just to fail at the dispatch stage, i.e. have no 128 case to match in `x86_execute`. But in any case, the critical thing is we aren't just allowing the bogus encoding, so that seems fine.

jargh (approved):
All looks good, thanks!

PR #291, opened 2025-09-19
Add AVX2 instructions needed for ML-KEM code

jargh:
This adds all the remaining AVX2 instruction models needed to load and reason about the x86 native code in the mlkem-native repository. (The `rej_uniform_asm.S` function needs some additional non-AVX2 instructions, but all other functions can be handled with these.)

```
        VMOVDQU
        VPOR
        VPSLLW
        VPSLLD
        VPSRLD
        VPBLENDW
        VPERMD
        VPSHUFB
```

This also makes some fixes and improvements to the existing AVX2 decoding infrastructure:

- `read_VEX` now correctly analyzes the implicit prefix parts in the case of a 2-byte VEX prefix.

- decoding is predicated by correct/standard implicit prefixes and where appropriate (for "`.W0`" or "`.W1`" instructions), the W field.


x86/proofs/decode.ml
```diff
@@ -595,99 +595,162 @@ let decode_aux = new_definition `!pfxs rex l. decode_aux pfxs rex l =
...
+           | (T, Rep0, SG0) -> SOME (VPERMD (mmreg reg sz) (mmreg v sz) (simd_of_RM sz rm),l)
+           | _ -> NONE)
         | [0x58:8] ->
           let sz = vexL_size L in
           (read_ModRM rex l >>= \((reg,rm),l).
-            SOME (VPBROADCASTD (mmreg reg sz) (simd_of_RM (if sz = Lower_256 then Lower_128 else sz) rm),l))
+           if rex_W rex then NONE else
+           match pfxs with
+           | (T, Rep0, SG0) -> SOME (VPBROADCASTD (mmreg reg sz) (simd_of_RM (if sz = Lower_256 then Lower_128 else sz) rm),l)
+           | _ -> NONE)
```
pennyannn, line 623: Could `if sz = Lower_256 then Lower_128 else sz` be simplified to `Lower_128` since it seems like we are only working with ymm and xmm registers?

x86/proofs/decode.ml
```diff
@@ -595,99 +595,162 @@ let decode_aux = new_definition `!pfxs rex l. decode_aux pfxs rex l =
...
         | [0x59:8] ->
           let sz = vexL_size L in
           (read_ModRM rex l >>= \((reg,rm),l).
-            SOME (VPBROADCASTQ (mmreg reg sz) (simd_of_RM (if sz = Lower_256 then Lower_128 else sz) rm),l))
+           if rex_W rex then NONE else
+           match pfxs with
+           | (T, Rep0, SG0) -> SOME (VPBROADCASTQ (mmreg reg sz) (simd_of_RM (if sz = Lower_256 then Lower_128 else sz) rm),l)
```
pennyannn, line 629: Similar here

x86/proofs/decode.ml
```diff
@@ -595,99 +595,162 @@ let decode_aux = new_definition `!pfxs rex l. decode_aux pfxs rex l =
...
+           | (T, Rep0, SG0) -> SOME (VPMULLD (mmreg reg sz) (mmreg v sz) (simd_of_RM sz rm),l)
+           | _ -> NONE)
         | [0xf6:8] ->
           let sz = op_size_W rex T pfxs in
           read_ModRM_operand rex sz l >>= \((reg,rm),l).
           (match pfxs with
-          | (F, RepNZ, SG0) ->
-            SOME (MULX4 (reg, %(Gpr v sz)) (%(Gpr (word 2) sz), rm), l)
-          | _ -> NONE)
+           | (F, RepNZ, SG0) ->
```
pennyannn, line 641: Opcode for this instruction: VEX.LZ.F2.0F38.W1 F6 /r 
LZ - If VEX.LZ is present in the opcode column: The VEX.L must be encoded to be 0B, an #UD occurs if VEX.L is not zero. Should we check for this?

pennyannn (commented):
It's great that the W0/W1, and prefix issues are addressed!

Add instructions that are DOIT (https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/resources/data-operand-independent-timing-instructions.html) to the allowed_asm file.

jargh:
Oh, and for no particularly strong reason except a vague sense of tidiness I re-sorted the `allowed_asm` list, which means that the changes to that file look more dramatic than they are. Modulo order, I'm just adding the new instructions as suggested.

PR #292, opened 2025-09-19
Integer instructions needed for ML-KEM code

jargh:
This adds the POPCNT and PEXT instructions.


x86/proofs/decode.ml
```diff
@@ -611,6 +617,13 @@ let decode_aux = new_definition `!pfxs rex l. decode_aux pfxs rex l =
           let sz = vexL_size L in
           (read_ModRM rex l >>= \((reg,rm),l).
             SOME (VPMULLD (mmreg reg sz) (mmreg v sz) (simd_of_RM sz rm),l))
+        | [0xf5:8] ->
+          let sz = op_size_W rex T pfxs in
+          read_ModRM_operand rex sz l >>= \((reg,rm),l).
+          (match pfxs with
```
pennyannn, line 623: Instruction opcode: VEX.LZ.F3.0F38.W1 F5 /r
If VEX.LZ is present in the opcode column: The VEX.L must be encoded to be 0B, an #UD occurs if VEX.L is not zero.
Should we add a check for `LZ` and `W1`?

pennyannn (commented):
These two instructions are not in DOIT table, so we don't have to add them.

jargh:
The last commit factors out the PEXT semantics into a custom word operation `word_condense` with its own evaluation conversion. The actual semantics is the same as before, but the expressions coming out of symbolic simulation will look less clumsy.

pennyannn (approved):
LGTM and the way `word_condense` is defined is so cool.

PR #293, opened 2025-09-20
SSE instructions needed for ML-KEM code

jargh:
This adds all the SSE or SSE-type instructions needed to load
and reason about the ML-KEM code, plus a couple of others
that were added for regularity:

```
        MOVQ
        PCMPGTW
        PSRLW
        PSHUFB
        PBLENDW
        PINSRD
        PINSRQ
        PMOVMSKB
```


x86/proofs/x86.ml
```diff
@@ -1064,11 +1120,17 @@ let x86_PSHUFD = new_definition
 let x86_PSRAD = new_definition
   `x86_PSRAD dest imm8 s =
     let d = read dest s in
-    let count_src = val (read imm8 s) in
-    let count = if count_src > 31 then 32 else count_src in
+    let count = val (read imm8 s) in
```
pennyannn, line 1123: Thanks for fixing the redundant line :)

jargh:
Thanks, I've added the new instructions to the list. It all seems good including an offline `make proofs`.

PR #295, opened 2025-09-22
fix f-string syntaxerror in collect-signatures.py

kings-crown:
There was a f-string SyntaxError across Python versions. 

Changed the inner quotes to single quotes
before:
print(f"  - {argname}: {"const " if isconst else ""}{argtype}")

after:
print(f"  - {argname}: {'const ' if isconst else ''}{argtype}")


aqjune-aws (approved):
Thanks for this fix for lower versions of Python 3 !

PR #297, opened 2025-10-06
ML-KEM poly_reduce proof for x86 AVX2 implementation

dkostic:
*Issue #, if available:*
N/A

*Description of changes:*
This change adds the implementation of ML-KEM poly_reduce
the implementation is the same as the one in mlkem-native
repository at https://github.com/pq-code-package/mlkem-native.


jargh (commented):
This is pretty close to being ready to merge - the code all seems to work well and the core proof is good. A few suggestions, the first two pretty minor:
* Let's add an explicit alignment to the buffer used in `test_mlkem_reduce` so it makes sure that the aligned loads are safe, e.g.
`int16_t a[256], b[256] __attribute__((aligned(32)));`
* The comment banner of the function should have the corresponding x86 ABI rather than the cut-and-pasted ARM one:
`// Standard x86-64 ABI: RDI = a`
`// Microsoft x64 ABI:   RCX = a`

* Most important would be to have the four variants of the main subroutine correctness (with/without IBT, with/without Windows), which should be largely automatic from the core theorem that you've proved. The corresponding `mldsa_reduce` proof may be a good guide here, but I can also help to fill this out - it's mainly routine boilerplate but it's a bit fiddly to get right until you are used to it.

jargh (changes requested):
Almost perfect: `make proofs` ran without a hitch and all my other tests went well, with just one small exception: I tried remaking the `x86_att` forms with `make clobber; make` in that directory, and it doesn't work on this file. I think there is a need to an additional pattern near the end of `attrofy.sed` (around line 134) that changes some sizes from q depending on the operands. But I'm too sleepy to figure it out :-)

PR #305, opened 2025-11-04
ML-KEM basemul_k2 proof for x86 AVX2 implementation

dkostic:
*Issue #, if available:*
N/A

*Description of changes:*
This change adds the implementation of ML-KEM basemul_k2
the implementation is the same as the one in mlkem-native
repository at https://github.com/pq-code-package/mlkem-native.

The new congbound rules, CONGBOUND_MONTMUL_X86 and
CONGBOUND_WORD_NEG, were implemented by @jargh .


dkostic:
> Almost perfect now, with all the proofs working fine, thanks!
> 
> Just one thing: the `x86_att` form is out of step with the `x86` one, and doing `make clobber; make` doesn't work because the sed script doesn't handle the form of relative addresses `[42+rsi]`, only `[rsi+42]`. Maybe the path of least resistance is just to switch the syntax for relative addresses round in the original, since the code is the same anyway and it fits the s2n-bignum style better. I've attached a patch with the effective addresses rewritten, and confirmed that it works. [basemul_k2.patch](https://github.com/user-attachments/files/23400088/basemul_k2.patch)

I patched the sed script to handle those as well. The issue was that the x86 code contained a few tabs and that didn't work well with attrofy. I fixed that in the latest update.

jargh:
Ah, OK, I see, thanks! It's nice that the script can handle a bit more, anyway. It all looks fine to me now - by the way I like the way you factored out the different x86 element ordering into separate reordering functions in the testing framework. I'll just rerun the proof and assuming no problems arise, will approve and merge it shortly.

jargh (approved):

PR #306, opened 2025-11-04
Update VMOVQ

manastasova:
*Issue #, if available:*

*Description of changes:*
Update x86 decoding for: VMOVQ.


jargh:
All the decoding setup seems perfect now - thanks for figuring it all out properly! I even tried it on some non-standard encodings like this
```
        .byte 0x66, 0x48, 0x0f, 0x7e, 0x0c, 0x24 // movq  QWORD PTR [rsp], xmm1 uniform encoding
        .byte 0x66, 0x41, 0x0f, 0xd6, 0xc7      // movq xmm15, xmm0, encoded with 66
```
and they all work as hoped. The remaining issues seem to be around the final dispatch and simulation. There are actually two distinct issues:
* We need some extra cases in the dispatching definitions and in the shortcut theorem `OPERAND_SIZE_CASES` to allow for the (128,128) operand size combination; this was the suggestion in our discussion yesterday.
* To handle the case of xmm -> xmm operations, we need to tweak the definition of `x86_[V]MOVQ` to explicitly force a 64-bit intermediate value, otherwise it's copying the full 128 bits; previously this didn't matter because either the source or destination was always 64-bit. (Stylistically there could be a case for doing the same with `x86_[V]MOVD` and 32 bits, but there are no paths where the current definition is a problem.)

I believe that the attached patch includes all the changes needed (?)
[movq.patch](https://github.com/user-attachments/files/23541413/movq.patch)

PR #320, opened 2025-12-15
x86 uniform rejection sampling for ML-KEM, MOVSB and related additions

jargh:
This adds the x86 rejection sampling code from `mlkem-native` with a couple of minor changes (see comments in the code), and its correctness proof. The correctness proof is in effect a simplified version of the ARM proof, with the same kind of underlying logic but much simpler loop structure. 

The more complicated side of this PR is the setup of infrastructure to support a single instruction `REP MOVSB`, which is used as the last step of the code to copy from the temporary buffer on the stack to the output. This is rather different from all the x86 instructions formalized so far in that it manipulates arbitrarily-sized buffers of memory, not just individual chunks 8, 64, 256 bits or whatever. Moreover the semantics is explicitly sequential, and handling the case of overlapping input and output buffers needs to take account of this. An auxiliary function `x86_movsb` is defined recursively, and for simulation this is factored into an equivalent using `x86_stringcopy` with bounded input and output buffers.

A second level of additional work is needed because repetition inside `REP MOVSB` is controlled by the `DF` flag, not previously part of the x86 formalization (this determines whether iteration is downward or upward). We add this to the model and the simulator, as well as the two instructions for explicitly setting and clearing it (`CLD`, `STD`). Finally, all the `MOVSB` and `REP MOVSB` semantics is checked in the simulator for sizes up to 7 bytes, which should be enough to exercise the various corner cases.

### DIFF
--- a/x86_att/mlkem/mlkem_rej_uniform_VARIABLE_TIME.S
+++ b/x86_att/mlkem/mlkem_rej_uniform_VARIABLE_TIME.S
@@ -12,7 +12,7 @@
 //                      const uint8_t *table);
 //
 // Interprets the input buffer as packed 12-bit numbers with a length of
-// buflen bytes, assumed to be a multiple of 24. Fills the output array
+// buflen bytes, assumed to be a multiple of 12. Fills the output array
 // with those numbers from the packed buffer that are < 3329, in the order
 // of appearance, returning the total number of entries written, with a
 // maximum of 256. The table argument is a specific precomputed table of


x86/mlkem/mlkem_rej_uniform_VARIABLE_TIME.S
```diff
@@ -0,0 +1,128 @@
...
+// Uniform rejection sampling for ML-KEM
+// Inputs *buf (unsigned bytes), buflen, table (unsigned bytes); output r[256] (signed 16-bit words), return
+//
+// extern uint64_t mlkem_rej_uniform_VARIABLE_TIME
+//                     (int16_t r[S2N_BIGNUM_STATIC 256],
+//                      const uint8_t *buf,uint64_t buflen,
+//                      const uint8_t *table);
+//
+// Interprets the input buffer as packed 12-bit numbers with a length of
+// buflen bytes, assumed to be a multiple of 24. Fills the output array
```
aqjune-aws: I found that the spec `MLKEM_REJ_UNIFORM_WINDOWS_SUBROUTINE_CORRECT` is more generous in the sense that `buflen` is okay to be a multiple of 12. Should we change this statement to be a multiple of 24? Or I guess the current wording is still fine because it is consistent with what Arm's mlkem_rej_uniform_VARIABLE_TIME says.

aqjune-aws (approved):
I confirmed that the cosimulator ran fine after `MOVSB` is only left. The decoding and semantics of `MOVSB` looks fine. I also (somewhat heuristically) checked that the precondition about `read DF s` in `MLKEM_REJ_UNIFORM_SUBROUTINE_CORRECT` is necessary for the proof, by confirming that commenting it out and running the proof failed. specifications.txt is also correctly updated. :)

PR #321, opened 2025-12-16
Correct x86 ML-KEM inverse NTT output bound

mkannwischer:
While integrating the x86 ML-KEM inverse NTT proof into mlkem-native (more specifically when adding the corresponding CBMC contract and trying to prove its adherence to the API contracts), we noticed that the output bound in the HOL-Light specification is off by one:
mlkem-native requires < 26632 (8q), while the HOL-Light specification states <= 26632.
This commit corrects the specification. No change to the proof is required.

 - See https://github.com/pq-code-package/mlkem-native/pull/1394

### DIFF
--- a/x86/proofs/mlkem_intt.ml
+++ b/x86/proofs/mlkem_intt.ml
@@ -1104,7 +1104,7 @@ let MLKEM_INTT_CORRECT = prove
                         ==> let zi =
                       read(memory :> bytes16(word_add a (word(2 * i)))) s in
                       (ival zi == avx2_inverse_ntt (ival o x) i) (mod &3329) /\
-                      abs(ival zi) <= &26632))
+                      abs(ival zi) <= &26631))
           (MAYCHANGE_REGS_AND_FLAGS_PERMITTED_BY_ABI  ,,
            MAYCHANGE [ZMM0; ZMM1; ZMM2; ZMM3; ZMM4; ZMM5; ZMM6; ZMM7; ZMM8;
                       ZMM9; ZMM10; ZMM11; ZMM12; ZMM13; ZMM14; ZMM15] ,,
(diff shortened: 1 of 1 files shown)


mkannwischer:
> Concretely, if I run the proof up to the splitting up into subgoals and application of `CONGBOUND_RULE`, then do this to examine the automatically produced bounds:
> 
> ```
> let gls = ((fun (_,l,_) -> l) (hd(!current_goalstack)));;
> 
> let lowerbounds = map (minus_num o dest_intconst o lhand o lhand o rand o lhand o snd) gls
> and upperbounds = map (dest_intconst o rand o rand o rand o lhand o snd) gls in
> end_itlist max_num (lowerbounds @ upperbounds);;
> ```
> 
> it looks as if we could justify a bound of <= 14939 (about 4.5q). That's not necessarily absolutely sharp, but I suspect it's not far off. Would it be beneficial to call that out in the spec? Or is it better just to keep the interface consistent? I don't have any very clear view either way.

Thanks for looking into this!

It makes sense that you can prove a much smaller bound because Montgomery multiplication has smaller output bounds than Barrett multiplication. In fact, we could probably revisit the number of reductions in that code - the current bounds reasoning uses q as the output bound for the Montgomery multiplications, but the actual bound for signed Montgomery multiplication with a canonical constant is 3/4q. @hanno-becker, do you think it would make sense to revisit the additional reduction you added to the pqcrystals code? It would be nice if we can prove the current bound without the additional reduction.

Smaller bounds are certainly not a problem for mlkem-native, but we will likely not change the API contracts/subsequent code. To me, there is value in having a smaller bound both to compare to other work verifying similiar code and for using this code in other contexts.

jargh:
I'm planning to just merge this as is since it does qualify as a bugfix. But I'll also open up an issue to make sure we revisit the question of whether to sharpen any input or output bounds in the proofs (or even decide we can safely optimize the code).

PR #332, opened 2025-12-31
Prove safety of P-521, X25519, Ed25519 and Karatsuba Mults

aqjune-aws:
This patch proves safety of P-521, X25519, Ed25519 and Karatsuba multiplications
that are used in AWS-LC.

The detailed list is:

1. P-521

Point-level operations (both arm and x86):

* p521_jscalarmul(_alt)
* p521_jdouble(_alt)

For x86, the following extra functions are verified:

* p521_jadd(_alt)
* bignum_mod_p521_9
* bignum_mod_n521_9(_alt)

2. X25519

Point-level operations (x86):
* curve25519_x25519(_alt)
* curve25519_x25519base(_alt)

Point-level operations (arm):
* curve25519_x25519_byte(_alt)
* curve25519_x25519base_bytes(_alt)

Modular operations (both arm and x86):
* bignum_inv_p25519
* bignum_mod_n25519

3. Ed25519

Point-level operations (both arm and x86):
* edwards25519_scalarmuldouble(_alt)
* edwards25519_scalarmulbase(_alt)

Decode (both arm and x86):
* edwards25519_decode(_alt)

4. Karatsuba Mult (Arm only)

* bignum_k{mul,sqr}_16_32
* bignum_k{mul,sqr}_32_64

And additional updates are:

* Reorder blocks in the assembly implementation of x86 bignum_mod_n25519 so that WINDOWS_X86_WRAP_STACK_TAC can be used.
* Add safety_print_log boolean reference flag for debugging
* Update ARM_ADD_RETURN_NOSTACK_TAC.
* Minor updates to the x86 add/wrap_stack fns to avoid repetition
* Add CANONICALIZE_UNIFY_F_EVENTS_TAC and make DISCHARGE_SAFETY_PROPERTY_TAC more powerful using it
* Fix a bug in gen_mk_safety_spec and mk_noibt_subroutine_safe_spec

---

Another small commit is faster x86 symbolic simulation by preprocessing x86_execute.

This patch makes x86 symbolic simulation slightly faster by preprocessing x86_execute for possible instruction kinds.
This does not significantly reduce the total running time of proofs, but reduces ~0.1 sec for each step, which is kind of satisfying.

PR #334, opened 2025-12-31
Add tutorial for safety proofs

aqjune-aws:
This patch adds Arm and x86 tutorials for safety proofs. :)

This is built on top of #332, and only the last two commits of this pull request are the new ones.


jargh (approved):
All looks great, thank you! I've run and tested everything and have also made use of the tutorial to try out a safety proof for myself. I will approve and merge :-)

PR #343, opened 2026-02-02
x86: Add VPMULHRSW (Packed Multiply High with Round and Scale)

mkannwischer:
Add support for the VPMULHRSW AVX/AVX2 instruction, which computes a rounded multiply-high for each 16-bit lane:
dst[i] = ((src1[i] * src2[i]) >> 14 + 1) >> 1

This instruction is used in ML-KEM compress/decompress AVX2 assembly.

*Issue #, if available:*

*Description of changes:*

I closely followed the preexisting VPMULHW, but I'm not super familar with x86, so please take a look if the change makes sense. I ran the sematests for a while on my x86 and it all seems to work fine.

Context: In [mlkem-native](https://github.com/pq-code-package/mlkem-native), @hanno-becker and I are currently working on converting the compress/decompress intrinsics to assembly and proving them correct. The first step is the proof for `poly_decompress_d4` which has just been completed in https://github.com/pq-code-package/mlkem-native/pull/1533. For that we require the VPMULHRSW instruction. After that there are going to be 7 more proofs and we will likely require another 11 instructions. I'm hoping to complete all that by RWC in March.


mkannwischer:
We have completed the proofs for all decompression fucntions. I'll open a new PR shortly with all the required instructions.

PR #344, opened 2026-02-04
 Arm64: Add vector `ABS`, `CMGE`, and `UMAXV` instructions

mkannwischer:
*Issue #, if available:*

*Description of changes:*

This PR adds 3 Neon instructions `abs`, `cmge`, `umaxv` that I need for the proof of poly_chknorm in mldsa-native - I have just completed that proof [here](https://github.com/pq-code-package/mldsa-native/pull/942). I can open a PR with that proof as well after it has been reviewed if that is of interest for you.

I tried to follow the modelling of existing instructions, but please let me know if anything can be improved! 
- `cmge` closely follows `cmhi`
- `abs` closely follows `add`
- `umaxv`  closely follows `uaddlv`

The `umaxv` modelling is quite ugly, please let me know if there is a better way to express it. `uaddlv` is using `nsum` which is much cleaner, but there is no `nmax` I believe.

I ran the `sematest` for all 3 instructions for a while.

### DIFF
--- a/arm/proofs/instruction.ml
+++ b/arm/proofs/instruction.ml
@@ -1714,42 +1714,54 @@ let arm_UADDLV = define
                     (\i. val(word_subword n (esize*i,esize):int128)) in
          (Rd := (word d:128 word)) s`;;
 
+let NEUTRAL_MAX = prove
+ (`neutral MAX = 0`,
+  REWRITE_TAC[neutral] THEN MATCH_MP_TAC SELECT_UNIQUE THEN
+  REWRITE_TAC[] THEN
+  GEN_TAC THEN EQ_TAC THENL [ALL_TAC; ARITH_TAC] THEN
+  DISCH_THEN(MP_TAC o SPEC `0`) THEN ARITH_TAC);;
+
+let MONOIDAL_MAX = prove
+ (`monoidal MAX`,
+  REWRITE_TAC[monoidal; NEUTRAL_MAX] THEN ARITH_TAC);;
+
+let ITERATE_MAX_CLAUSES =
+  REWRITE_RULE[NEUTRAL_MAX] (MATCH_MP ITERATE_CLAUSES MONOIDAL_MAX);;
+
+let EXPAND_ITERATE_MAX_CONV =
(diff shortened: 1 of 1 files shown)


jargh:
This all seems to be fully functionally correct - as well as manual inspection I've run various tests and checks and everything works well. But indeed it is a bit unfortunate to have the core `arm_UMAXV` so verbose since it works against surveyability of the specification. It would be possibly to do something analogous to `arm_UADDLV` by replacing the `nsum` in its definition by `iterate MAX` (here `iterate` is a more general "iterate a commutative-monoid-like operation over a set" that is used to define `nsum` and others). However there is no supporting infrastructure for this, nothing as convenient as `EXPAND_NSUM_CONV` so one would need a bit of extra work there. Still I think it's worth it.

mkannwischer:
> For example you could set up the infrastructure with
> 
> ```
> let NEUTRAL_MAX = prove
>  (`neutral MAX = 0`,
>   REWRITE_TAC[neutral] THEN MATCH_MP_TAC SELECT_UNIQUE THEN
>   REWRITE_TAC[] THEN
>   GEN_TAC THEN EQ_TAC THENL [ALL_TAC; ARITH_TAC] THEN
>   DISCH_THEN(MP_TAC o SPEC `0`) THEN ARITH_TAC);;
> 
> let MONOIDAL_MAX = prove
>  (`monoidal MAX`,
>   REWRITE_TAC[monoidal; NEUTRAL_MAX] THEN ARITH_TAC);;
> 
> let ITERATE_MAX_CLAUSES =
>   REWRITE_RULE[NEUTRAL_MAX] (MATCH_MP ITERATE_CLAUSES MONOIDAL_MAX);;
> ```
> 
> I can help with setting up the evaluation conversion if you like.

Thank you @jargh for these hints. With that, I managed to get something much cleaner to work with some help from Claude Opus 4.6. I updated my proof accordingly.

jargh:
Sorry for the delay - this all looks great now, thanks! Would you mind rebasing for the last couple of merges and then I'll be ready to go ahead and approve.

PR #346, opened 2026-02-05
x86: Add missing instructions for ML-KEM decompressions

mkannwischer:
*Issue #, if available:*

*Description of changes:*

This PR adds 5 missing AVX2 instructions that we require for the ML-KEM decompressions. With those @hanno-becker was able to prove correctness of the 4 decompression functions for d=4,5,10,11 (see [this PR](https://github.com/pq-code-package/mlkem-native/pull/1543)). Please see the individual commits for details. The instructions for the 4 compression functions will be added separately. 

I ran the sematests for all instructions for a while, that plus the completed proofs give us some confidence that the instruction models are actually correct, but please take a look.

I have a question on how the sematest works: My understanding is that the sematest runs the instruction natively with random data and then compares the result to the HOL-Light model. However, the variable bit shifting only really makes sense if the shift distances are smaller than the element size, otherwise the output is always going to be zero. Picking random shift distances will result in zero with very high probability. Is my understanding correct and is there a way to test this better?


dkostic (commented):
I checked out the PR locally, modified `simulator.ml` to remove test iclasses for all but the instructions that were added in this PR, and run `make sematest`:
```
...
Finished (time limit: 2400.000000s, tested register-only: 1120, general memory: 141, special memory: 120, total: 1381)
Running time: 2606.000000 sec, Start unixtime: 1771886554.000000, End unixtime: 1771889160.000000
- Simulator 32 finished successfully
```

PR #349, opened 2026-02-12
x86: Add missing instructions for ML-KEM compressions

mkannwischer:
*Issue #, if available:*

*Description of changes:*

- Depends on https://github.com/awslabs/s2n-bignum/pull/346

This PR adds 7 additional missing AVX2 instructions that we require for the ML-KEM compressions. With those @hanno-becker was able to prove correctness of the 4 compression functions for d=4,5,10,11 (see https://github.com/pq-code-package/mlkem-native/pull/1545/). The 5 missing instructions for the  4 decompression functions proven in https://github.com/pq-code-package/mlkem-native/pull/1543 were already added as a part of https://github.com/awslabs/s2n-bignum/pull/346.

I ran the sematests for all instructions for a while, that plus the completed proofs give us some confidence that the instruction models are actually correct, but please take a look.

This completes the proofs of the arithmetic x86 backend of ML-KEM. 🎉🥳


jargh (commented):
All this looks great and all my tests are fine. It would be nice if you could also add a few more simulator test cases to sanity-check (1) memory versions of the new instructions where applicable, and (2) oversized immediates to check masking or otherwise where applicable. This patch gives a few examples that could be added. I don't consider this critical since I've already done so locally and confirmed they all seem fine, so if you are in a rush I can just approve anyway :-)

[349.patch](https://github.com/user-attachments/files/25595485/349.patch)

PR #352, opened 2026-02-17
ML-DSA Forward NTT proof for Arm

dkostic:
This change adds the Arm implementation of ML-DSA Forward NTT
the implementation is the same as the one in mldsa-native
repository at https://github.com/pq-code-package/mldsa-native.

The proof was developed with the help of Anthropic Opus 4.6.


jargh:
All looks great, thank you, and all proofs and tests look fine. Just two minor requests:
* Could you rebase for the current mainline, which has had quite a few merges since this PR
* Could you also update/regenerate the `include/s2n-bignum-c89.h` header?

The latter can be done by `make clobber; make` in the `include` directory, and it might actually fix a few other incompatibilities there too(?)

dkostic:
> All looks great, thank you, and all proofs and tests look fine. Just two minor requests:
> 
> * Could you rebase for the current mainline, which has had quite a few merges since this PR
> * Could you also update/regenerate the `include/s2n-bignum-c89.h` header?
> 
> The latter can be done by `make clobber; make` in the `include` directory, and it might actually fix a few other incompatibilities there too(?)

Done! Thanks for the feedback!

PR #353, opened 2026-02-17
ML-DSA Inverse NTT proof for Arm

dkostic:
This change adds the Arm implementation of ML-DSA Inverse NTT
the implementation is the same as the one in mldsa-native
repository at https://github.com/pq-code-package/mldsa-native.

The proof was developed with the help of Anthropic Opus 4.6.


jargh (approved):
All looks great - the code structure is fine, the proofs all run and the bounds are the same as the x86 version and match the API assumptions for the mldsa-native front end. I will merge.

PR #354, opened 2026-02-17
Optimized ALT AVX2 Keccak X4 implementation, spec, and proof

manastasova:
**Add Optimized AVX2-optimized x4 batched Keccak implementation**

This PR introduces a Optimized Alt SIMD Keccak implementation using AVX2 ISA that processes 4 hash computations in parallel.

Optimizations against https://github.com/awslabs/s2n-bignum/pull/330:
- Register allocation: Reduced stack spills through better register utilization in the AVX2 Keccak permutation, lowering the stack frame
- Formal verification: HOL Light proof of correctness `SHA3_KECCAK4_F1600_ALT_SUBROUTINE_CORRECT`
- Performance Results: `cd s2n-bignum/benchmark && make go`

### DIFF
--- a/x86/sha3/sha3_keccak4_f1600_alt.S
+++ b/x86/sha3/sha3_keccak4_f1600_alt.S
@@ -658,7 +658,7 @@ Lsha3_keccak4_f1600_alt:
      vmovhpd [rdi+0x188],xmm3
      vmovq [rdi+0x250],xmm15
      vmovhpd [rdi+0x318],xmm15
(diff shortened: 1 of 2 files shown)


jargh (approved):
All code, tests and proofs look good, thanks! I also see a significant performance boost on several platforms I've tried relative to the earlier 4x version. It's a bit unfortunate to have the more complex API constant table requirements, but we can unify those eventually when the support for rodata on x86 has been finalized.

PR #372, opened 2026-03-13
ML-DSA Aarch64 HOL-Light proof poly_use_hint_32

jakemas:
## Summary

With help, and love, from Claude.

Add formal verification proof for `mldsa_poly_use_hint_32` (AArch64 NEON), establishing full functional correctness: the assembly output equals `word(mldsa_use_hint_32_spec(val(a), val(h)))` for each element, matching the FIPS 204 UseHint specification for GAMMA2 = (Q-1)/32 (ML-DSA security levels 3 and 5, l=5,7).

The companion `_88` proof (security level 2, l=4) is in PR #375.

## What the proof establishes

The proof verifies that hand-optimized NEON assembly — Barrett multiplication with magic constants, SIMD bit tricks, branchless conditional masking — produces exactly the same mathematical function as the FIPS 204 spec, for every valid input (8.4M x 2 input pairs).

## Proof architecture

**Barrett equivalence** (`BARRETT_EQUIV`): The assembly computes `a1` as `((2 * a * 1074791425) / 2^32 + 131072) / 262144`. The spec computes it as `((a + 127) / 128 * 1025 + 2097152) / 4194304`. These are completely different formulas. The proof establishes equivalence for all `a < Q` via 17-case analysis (one per output value 0..16), using `DIV_MONO` to sandwich both formulas to the same constant in each output interval.

**Element correctness** (`ELEMENT_CORRECT`): Proves `val(asm_result) = spec(val(a), val(h))` for a single element. Splits into wrap (a1=16) and nowrap (a1<=15) cases. Uses bridge lemmas for native-mode `real_gt`/`int_gt` compatibility, and `DIV_MONO`-based upper bounds for the remainder.

**SIMD endgame** (`MLDSA_USE_HINT_CORRECT`): After 1006-step ARM symbolic execution, pushes `word_subword` through SIMD operations to extract per-element 32-bit results, then matches each against the element correctness theorem.

**Shared helpers**: `WORD_AND_ONES_32`, `WORD_MUL_1_32`, `REAL_INT_GT_BRIDGE`, `REAL_INT_GT_BRIDGE_POS` moved to `common/mlkem_mldsa.ml` for reuse by both `_32` and `_88` proofs. Unused intermediate lemmas from development removed.

## Changes

- `arm/proofs/mldsa_poly_use_hint_32.ml` — full proof with functional correctness postcondition, constant-time and memory safety
- `arm/mldsa/mldsa_poly_use_hint_32.S` — assembly source
- `common/mlkem_mldsa.ml` — shared helper lemmas
- Supporting changes to Makefile, subroutine signatures, specifications, decode/instruction tables

## Testing

Proof verified end-to-end in HOL Light (both interpreted MCP and native compilation CI). Also verified in the [mldsa-native](https://github.com/pq-code-package/mldsa-native/pull/1037) fork which runs native-mode CI. The `_32` proof takes approximately 16 minutes in CI.


jargh:
I'm still thinking about the interesting aspects of this PR, but I noticed one oddity: `arm/proofs/subroutine_signatures.ml` seems to have an erroneous duplication at lines 4575-4578 (?)

jakemas:
> I'm still thinking about the interesting aspects of this PR, but I noticed one oddity: `arm/proofs/subroutine_signatures.ml` seems to have an erroneous duplication at lines 4575-4578 (?)

Hanno suggested a re-write to make the format in line with other proof styles: https://github.com/pq-code-package/mldsa-native/pull/1037#issuecomment-4363057304 which I will do

jakemas:
> I'm still thinking about the interesting aspects of this PR, but I noticed one oddity: `arm/proofs/subroutine_signatures.ml` seems to have an erroneous duplication at lines 4575-4578 (?)

made some changes in [3956e68](https://github.com/awslabs/s2n-bignum/commit/3956e68fc77b618a5dfc3cda1d27a425e7e9a96f)

jargh (approved):
All looks good now, thanks! I like the final form of the specification.

PR #375, opened 2026-03-13
ML-DSA Aarch64 HOL-Light proof poly_use_hint_88

jakemas:
## Summary

With help, and love, from Claude.

Add formal verification proof for `mldsa_poly_use_hint_88` (AArch64 NEON), establishing full functional correctness: the assembly output equals `word(mldsa_use_hint_88_spec(val(a), val(h)))` for each element, matching the FIPS 204 UseHint specification for GAMMA2 = (Q-1)/88 (ML-DSA security level 2, l=4).

This is the companion to the `_32` proof in PR #372.

## What the proof establishes

The proof verifies that hand-optimized NEON assembly — Barrett multiplication with magic constants, SIMD bit tricks, branchless conditional masking — produces exactly the same mathematical function as the FIPS 204 spec, for every valid input (8.4M x 2 input pairs).

## Proof architecture

**Barrett equivalence** (`BARRETT_EQUIV_88`): The assembly computes `a1` as `((2 * a * 1477838209) / 2^32 + 65536) / 131072`. The spec computes it as `((a + 127) / 128 * 11275 + 8388608) / 16777216`. These are completely different formulas. The proof establishes equivalence for all `a < Q` via 44-case analysis, using `DIV_MONO` to sandwich both formulas to the same constant in each output interval.

**Element correctness** (`ELEMENT_CORRECT_88`): Proves `val(asm_result) = spec(val(a), val(h))` for a single element. Splits into wrap (a1=44) and nowrap (a1<=43) cases. Uses bridge lemmas for native-mode `real_gt`/`int_gt` compatibility, BDD-based BITBLAST for word comparison bounds, and `DIV_MONO`-based upper bounds for the remainder.

**SIMD endgame** (`MLDSA_USE_HINT_88_CORRECT`): After 1006-step ARM symbolic execution, pushes `word_subword` through SIMD operations to extract per-element 32-bit results, then matches each against the element correctness theorem.

**Shared helpers**: `WORD_AND_ONES_32`, `WORD_MUL_1_32`, `REAL_INT_GT_BRIDGE`, `REAL_INT_GT_BRIDGE_POS` moved to `common/mlkem_mldsa.ml` for reuse by both `_32` and `_88` proofs.

## Changes

- `arm/proofs/mldsa_poly_use_hint_88.ml` — full proof with functional correctness postcondition, constant-time and memory safety
- `arm/mldsa/mldsa_poly_use_hint_88.S` — assembly source
- `common/mlkem_mldsa.ml` — shared helper lemmas
- Supporting changes to Makefile, subroutine signatures, specifications, decode/instruction tables

## Testing

Proof verified end-to-end in HOL Light (both interpreted MCP and native compilation CI). Also verified in the [mldsa-native](https://github.com/pq-code-package/mldsa-native/pull/1037) fork which runs native-mode CI.


jargh (approved):
All looks excellent now, thanks! I'll merge.

PR #378, opened 2026-03-18
ML-DSA aarch64 rejection sampling proof

dkostic:
*Description of changes:*
ML-DSA aarch64 rejection sampling proof.

This proof was written entirely by Claude Opus 4.6.


jakemas:
This is missing benchmarks/test.c, I've written them as part of https://github.com/awslabs/s2n-bignum/pull/401 so we coordinate a merge, I can always cherry pick from here. It would be good to get the instruction added in this PR decode additions merged in, as [other](https://github.com/pq-code-package/mldsa-native/pull/1040) proofs need it too. 

Other review, also missing specifictions.txt, s2n-bignum.h, collect-signatures.py, SAFE proof for memory safery (not constant time). Pointing claude at https://github.com/jakemas/s2n-bignum/blob/main/doc/s2n_bignum_pr_checklist.md could pick these off easily.

jargh:
The basic proof actually looks great, and that's the core of it. But I think we need the `..._SUBROUTINE_CORRECT` form, which hopefully is a routine/automatic wrapping here. It's also been in the queue a while so by now needs a rebase.

jargh:
All the essentials are fine, really; the failure is just because the new correctness theorems aren't entered in `arm/proofs/specifications.txt`. For consistency with the checks in `make proofs` it's probably better for now to use just `_SAFE` instead of `_MEMSAFE` for the memory safety proof, even though I agree the distinction is worth making. Also there's a bit of library duplication it'd be nice to cut. I think if you:
* Apply the attached patch
* Rebase for some other recent merges

it'll be perfectly ready to merge.
[pr378.patch](https://github.com/user-attachments/files/28923479/pr378.patch)

jargh:
I've added test and benchmarking support directly - we'll appropriately unify the respective setups when 401 is merged.

jargh (approved):
I believe this is all now solid and I'll merge

PR #381, opened 2026-03-25
ARM PMUL/PMULL/PMULL2 and x86 VPCLMULQDQ instructions via word_pmul

jargh:
*Issue #, if available:*

*Description of changes:*


sgmenda:
I was reading through https://github.com/nebeid/s2n-bignum/tree/ghash-polyval and [independently added](https://github.com/jrh13/hol-light/commit/b03c98d81d34e86623c4ee97fdf1e9befd3761fe) the same operation (as `word_clmul`) to HOL Light's `Library/words.ml`. not realizing the branch was based on this PR. 

Should `word_pmul` eventually live in `Library/words.ml` since it is not specific to s2n-bignum?

Also, it might be worth adopting this test vector I used as a load-time smoke test:
```
(* https://github.com/google/boringssl/blob/7f5a43945aab78fe4e71459ac4881ff9033d73d8/crypto/cipher/test/cipher_tests.txt#L338-L344 *)
let _ = prove
 (`word_clmul (word 0x0388DACE60B6A392F328C2B971B2FE78 : 128 word)
              (word 0x66E94BD4EF8A2C3B884CFA59CA342B2E : 128 word)
              : 256 word =
   word 0x009B5741881E078903A29B1A4B43CAF41FF4A78A706884B0473EFE75603871D0`,
  CONV_TAC(LAND_CONV WORD_CLMUL_CONV) THEN REFL_TAC);;
```

sgmenda (approved):
lgtm. checked with HOL Light REPL.

nitpick: can you update the PR description (-> commit message) to list the instructions added and mention that they use the new `word_pmul`?

jargh:
Thank you! Yes, I'll tweak the PR title and hence the commit message before merging.

jargh:
How does it look now?

jargh:
There does seem to be an issue with one or two of the x86 tests, e.g.
`[0xc4; 0xe3; 0x6d; 0x44; 0xcb; 0x11] (* Should be VPCLMULQDQ (%_% ymm1) (%_% ymm2) (%_% ymm3) (Imm8 (word 17)) *)`
Maybe the out-of-range immediate is actually an error here versus being masked or ignored.

jargh:
Oh, it seems that I may have just been running my local tests on hardware that doesn't support the full instruction (Intel Skylake). This shows up as a blank output on that machine, but on a more recent one:
```
$ grep -o vpclmulqdq /proc/cpuinfo | head -1
vpclmulqdq
```

jargh:
OK, decided it looks good. When we actually insert any code using the `ymm` forms of the x86 `VPCLMULQDQ` instruction we'll have to be careful in the testing and benchmarking to gate it on instruction availability.

PR #388, opened 2026-04-08
Add POLYVAL/GHASH reduction proofs and Karatsuba carry-less multiplication

nebeid:
*Description of changes:*

## POLYVAL reduction and Karatsuba carry-less multiplication infrastructure

Formal HOL Light proofs for core algebraic building blocks needed to verify AES-GCM GHASH/POLYVAL computation. These are standalone mathematical lemmas with no dependencies beyond `common/ghash.ml` (separate PR by John Harrison) and upstream `Library/words.ml`.

### New files

- **`common/polyval.ml`** -- Defines the POLYVAL polynomial `Q(x) = x^128 + x^127 + x^126 + x^121 + 1` (RFC 8452 / bit-reflected GHASH polynomial), equivalence modulo Q(x), and one-step reduction correctness. Also defines the `polyval_reduce_prop3` function implementing Gueron's Proposition 3 (see below for details) and proves its correctness as `POLYVAL_REDUCE_PROP3_CORRECT` (strategy: define a 256-bit quotient witness, prove the 256-bit word identity lane-by-lane via `BITBLAST_TAC`, then lift to polynomial level).

- **`common/karatsuba_pmul_proof.ml`** -- Proves that `word_pmul(a,b)` equals the Karatsuba decomposition using three half-size pmuls with XOR tidy-up. Two-part proof: schoolbook identity via `RING_RULE` (works in any commutative ring), then Karatsuba tidy-up via `WORD_BITWISE_RULE` (uses char-2 property `a XOR a = 0`).

### Proposition 3 (Gueron 2023)

Given a 256-bit carry-less product `T = D*x^192 + C*x^128 + B*x^64 + A` (where A, B, C, D are 64-bit polynomials), Proposition 3 computes `T * x^{-128} mod Q(x)` -- equivalently, the POLYVAL `dot(a, b) = a * b * x^{-128} mod Q(x)` for `T = word_pmul(a, b)` -- using only two 64-to-128-bit carry-less multiplications by a fixed constant `W = 0xC200000000000000` (so `W(x) = x^63 + x^62 + x^57`) plus XORs:

```
V = B XOR low64(W*A)
U = C XOR A XOR high64(W*A)
F = U XOR low64(W*V)
G = D XOR V XOR high64(W*V)
result = G * x^64 + F    (128-bit)
```

The proof shows that `result * x^128 == T (mod Q(x))`, equivalently `result == T * x^{-128} (mod Q(x))`, via the 256-bit identity `result*x^128 + T = Q(x) * quotient` where `quotient = A + (B XOR low64(W*A)) * x^64` is a 128-bit witness.

### References
- RFC 8452 (AES-GCM-SIV), Section 3 and Appendix A, https://datatracker.ietf.org/doc/html/rfc8452
- Gueron, "A New Interpretation for the GHASH Authenticator of AES-GCM", CSCML 2023, LNCS 13914, pp. 424-438, https://link.springer.com/chapter/10.1007/978-3-031-34671-2_30


common/polyval.ml
```diff
@@ -0,0 +1,273 @@
...
+needs "common/ghash.ml";;
+
+(* ------------------------------------------------------------------------- *)
+(* The Q(x) polynomial: x^128 + x^127 + x^126 + x^121 + 1                  *)
+(* This has the special form x^128 + W(x)*x^64 + 1 where                    *)
+(* W(x) = x^63 + x^62 + x^57 = 0xC200000000000000.                         *)
+(* ------------------------------------------------------------------------- *)
+
+let polyval_poly = define
+ `polyval_poly =
```
manastasova, line 19: Do we need irreducibility proof for the polyval polynomial similar to the ghash one?

common/polyval.ml
```diff
@@ -0,0 +1,273 @@
...
+   [`p IN ring_carrier bool_poly`; `p' IN ring_carrier bool_poly`;
+    `q IN ring_carrier bool_poly`; `q' IN ring_carrier bool_poly`] THEN
+  ASM_SIMP_TAC[cong; mod_polyval; RING_MUL] THEN
+  ASM_MESON_TAC[IN_RING_IDEAL_ADD; IN_RING_IDEAL_LMUL;
+                RING_IDEAL_IDEAL_GENERATED; RING_RULE
+   `ring_add r (ring_mul r q (ring_sub r p p'))
+               (ring_mul r p' (ring_sub r q q')) =
+    ring_sub r (ring_mul r p q) (ring_mul r p' q')`]);;
+
+(* ------------------------------------------------------------------------- *)
```
manastasova, line 130: np: Some comment align fixes needed

common/polyval.ml
```diff
@@ -0,0 +1,273 @@
...
+let polyval_reduce_step = new_definition
+ `polyval_reduce_step (x:N word) =
+  word_xor (word_subword x (0,128):N word)
+           (word_pmul (word_ushr x 128)
+                      (word 0xC2000000000000000000000000000001:N word))`;;
+
+let POLY_EQUIV_POLYVAL_REDUCE_STEP = prove
+ (`!x:N word.
+        (poly_of_word(polyval_reduce_step x) == poly_of_word x) (mod_polyval)`,
+  let lemma = prove
```
manastasova, line 145: Reused in ghash.ml:433-441 so maybe can be a separate lemma.

common/polyval.ml
```diff
@@ -0,0 +1,273 @@
...
+(* = 0xC2000000000000000000000000000001.                                     *)
+(* ------------------------------------------------------------------------- *)
+
+let polyval_reduce_step = new_definition
+ `polyval_reduce_step (x:N word) =
+  word_xor (word_subword x (0,128):N word)
+           (word_pmul (word_ushr x 128)
+                      (word 0xC2000000000000000000000000000001:N word))`;;
+
+let POLY_EQUIV_POLYVAL_REDUCE_STEP = prove
```
manastasova, line 142: the ghash.ml uses slightly different namings, e.g, `POLY_EQUIV_GHASH_REDUCE1` for the same theorems over the ghash polynomial. Should we try unifying them?

common/polyval.ml
```diff
@@ -0,0 +1,273 @@
...
+(*                                                                           *)
+(* Q(x) = x^128 + W(x)*x^64 + 1 where W(x) = x^63 + x^62 + x^57.         *)
+(*                                                                           *)
+(* Input T = D*x^192 + C*x^128 + B*x^64 + A (each limb 64 bits).           *)
+(* Phase 1: fold A using W*A, producing 192-bit T1.                          *)
+(* Phase 2: fold V (low 64 of T1) using W*V, producing 128-bit T2.          *)
+(* Result: T2 = T * x^{-128} mod Q(x).                                      *)
+(* ------------------------------------------------------------------------- *)
+
+let polyval_reduce_prop3 = define
```
manastasova, line 209: Should the spec be in the polyval_prop3_proof.ml?

common/polyval.ml
```diff
@@ -0,0 +1,273 @@
...
+(* Equivalently: result * x^128 ≡ T (mod Q(x)).                             *)
+(*                                                                           *)
+(* Proof idea (from Gueron):                                                 *)
+(*   T2 * x^128 = T + A*Q(x) + V*Q(x)*x^64                                 *)
+(* where A = low 64 bits of T, V = B XOR low64(W*A).                        *)
+(* Since A*Q + V*Q*x^64 is a multiple of Q(x), we have                      *)
+(*   T2 * x^128 ≡ T (mod Q(x)).                                             *)
+(* ------------------------------------------------------------------------- *)
+
+(* TODO: Formal proof of POLYVAL_REDUCE_PROP3_CORRECT:
```
manastasova: This looks to be already done

common/polyval_prop3_proof.ml
```diff
@@ -0,0 +1,280 @@
...
+needs "common/polyval.ml";;
+
+(* ------------------------------------------------------------------------- *)
+(* The quotient witness for ideal membership.                                *)
+(* For T = D*x^192 + C*x^128 + B*x^64 + A, the quotient is A + V*x^64     *)
+(* where V = B XOR low64(W*A).                                              *)
+(* ------------------------------------------------------------------------- *)
+
+let polyval_quotient = new_definition
+ `polyval_quotient (t:256 word) =
```
manastasova: Interesting the the output is saved as 256 bits, maybe it is useful for later.

common/polyval_prop3_proof.ml
```diff
@@ -0,0 +1,280 @@
...
+                                   (word_shl x 126))
+                         (word_shl x 121))
+              x`,
+  GEN_TAC THEN ONCE_REWRITE_TAC[WORD_PMUL_SYM] THEN REWRITE_TAC[PMUL_Q_256]);;
+
+(* ------------------------------------------------------------------------- *)
+(* Decompose 256-bit word equality into 4 x 64-bit lane equalities.         *)
+(* ------------------------------------------------------------------------- *)
+
+let WORD_EQ_LANES_256 = prove
```
manastasova: The proof seems too complex for that _simple_ goal. Can we use directly `BITBALST_RULE` OR `BITBALST_TAC`? Maybe performance will drop if we use it?

common/polyval_prop3_proof.ml
```diff
@@ -0,0 +1,280 @@
...
+  MATCH_MP_TAC(REWRITE_RULE[IMP_CONJ_ALT] LET_TRANS) THEN
+  MP_TAC(SPEC `t:256 word` QUOTIENT_CLZ) THEN
+  CONV_TAC(ONCE_DEPTH_CONV DIMINDEX_CONV) THEN ARITH_TAC);;
+
+(* ------------------------------------------------------------------------- *)
+(* Main theorem: Proposition 3 computes T * x^{-128} mod Q(x).              *)
+(* Equivalently: result * x^128 ≡ T (mod Q(x)).                             *)
+(* ------------------------------------------------------------------------- *)
+
+let POLYVAL_REDUCE_PROP3_CORRECT = prove
```
manastasova: How elegant!

common/karatsuba_pmul_proof.ml
```diff
@@ -0,0 +1,218 @@
...
+ (`!w:128 word. poly_of_word(word_zx w : 256 word) = poly_of_word w`,
+  GEN_TAC THEN MATCH_MP_TAC POLY_OF_WORD_ZX THEN
+  REWRITE_TAC[DIMINDEX_128; DIMINDEX_256] THEN ARITH_TAC);;
+
+let zx_64_128 = prove
+ (`!w:64 word. poly_of_word(word_zx w : 128 word) = poly_of_word w`,
+  GEN_TAC THEN MATCH_MP_TAC POLY_OF_WORD_ZX THEN
+  REWRITE_TAC[DIMINDEX_64; DIMINDEX_128] THEN ARITH_TAC);;
+
+let PMUL_DEG_TAC =
```
manastasova: np. if we have it parametrized `let PMUL_DEG_TAC_GEN dim_thm =`, (also working for 64*64=128-bits), the `shl_zx_64_to_poly` proof may be simplified a bit.

common/karatsuba_pmul_proof.ml
```diff
@@ -0,0 +1,218 @@
...
+(* ------------------------------------------------------------------------- *)
+
+let SCHOOLBOOK_RING = GEN_ALL(RING_RULE
+  `ring_mul r (ring_add r a (ring_mul r b x))
+              (ring_add r c (ring_mul r d x)) =
+   ring_add r (ring_mul r a c)
+     (ring_add r (ring_mul r (ring_add r (ring_mul r a d) (ring_mul r b c)) x)
+                  (ring_mul r (ring_mul r b d) (ring_mul r x x)))`);;
+
+let BOOL_POLY_ADD_ASSOC = prove(
```
manastasova: This one is not used anywhere. `PMUL_SCHOOLBOOK` seems to use `RING_ADD_ASSOC` directly.

common/polyval.ml
nebeid: @/tmp/body.txt

common/polyval_prop3_proof.ml
nebeid: @/tmp/body.txt

common/polyval.ml
nebeid: Yes. We already have it: `RING_IRREDUCIBLE_POLYVAL_POLYNOMIAL` in `common/ghash_spec.ml`, using `RABIN_IRREDUCIBILITY_SUFFICIENT` in the same style as `ghash.ml`. It doesn't ship in this PR because `ghash_spec.ml` depends on John's `common/ghash.ml`, so it will land in the follow-up PR once that is merged. One thing worth flagging: we keep the irreducibility proof for completeness (and for any future reasoning that needs field structure in GF(2^128)/Q), but the actual `polyval_dot` / Horner-unrolling proofs on that next PR do not use it -- all cancellations go through coprimality `gcd(Q, x^n) = 1` (which follows from `Q(0) = 1`), not through field-structure / invertibility.

common/polyval.ml
nebeid: Done in fd0bb73d: comment-header alignment normalized to column 79 across all three files, plus a few non-ASCII `===` symbols replaced with `==`.

common/polyval.ml
nebeid: Agreed; once `ghash.ml` is on the main branch we can lift this into a shared lemma (either in `polyval.ml` or, probably better, back in `ghash.ml` since the original is there). Deferring to the follow-up PR so we don't duplicate the change here.

common/polyval.ml
nebeid: Agreed. Easiest to do once `ghash.ml` is merged so we can rename in one place. I'll unify the naming in the follow-up PR along with extracting the shared lemma (see the other comment).

common/polyval.ml
nebeid: Done in fd0bb73d: `common/polyval_prop3_proof.ml` has been folded into `common/polyval.ml`, matching the co-location of definitions + correctness proofs already used in `common/ghash.ml`.

common/polyval.ml
nebeid: Correct, that was a stale TODO; removed in fd0bb73d as part of the file merge.

common/polyval_prop3_proof.ml
nebeid: Nice catch. Mathematically the quotient has degree < 128, so a `128 word` would carry the same information -- `QUOTIENT_HIGH_ZERO` / `QUOTIENT_CLZ` formalize exactly that -- and upstream HOL Light's `word_pmul : M word -> N word -> P word` is deliberately polymorphic in all three widths (jrh13/hol-light#166, to model different CPU variants), so nothing forces the 256-bit width. The reason we ended up at `256 word` here is that the surrounding machinery is stated at 256-bit uniformly: `PMUL_Q_256`, `PMUL_Q_256_SYM`, and the `BITBLAST_TAC` closing `WORD_IDENTITY_PROP3` all rewrite `word_pmul x Q` as a sum of `word_shl x k` on a 256-word `x`. Picking 256 for the quotient is the cheapest way to hit that uniform typing without inserting `word_zx` bridges. In other words, it's local type ergonomics rather than anything planned for reuse. Happy to tighten it to 128-bit in a follow-up if we find a call site that benefits.

common/polyval_prop3_proof.ml
nebeid: Good suggestion. Done in fd0bb73d: `WORD_EQ_LANES_256` now uses `BITBLAST_TAC` directly; the proof goes from ~24 lines to 1. Load time cost for the file is essentially unchanged (the manual proof was ~5.4s, `BITBLAST_TAC` is ~8.2s on this goal, but that's dominated by the subsequent Prop-3 lane BITBLASTs anyway).

common/polyval_prop3_proof.ml
nebeid: Thanks! RING_RULE does nearly all the work once the schoolbook identity is stated in a ring-generic form.

common/karatsuba_pmul_proof.ml
nebeid: Done in fd0bb73d: introduced `PMUL_DEG_TAC_GEN dim_out dim_in`, with `PMUL_DEG_TAC = PMUL_DEG_TAC_GEN DIMINDEX_256 DIMINDEX_128`. `shl_zx_64_to_poly` now ends with a single `PMUL_DEG_TAC_GEN DIMINDEX_128 DIMINDEX_64` application (was ~7 lines of manual bound discharge).

common/karatsuba_pmul_proof.ml
nebeid: Confirmed unused; removed in fd0bb73d. (`PMUL_SCHOOLBOOK` indeed just uses `RING_RULE` on a generic ring identity, which handles associativity internally.)

manastasova (approved):

jargh:
Thank you both, this is now great! One minute nit: the `POLY_VAR_POW_OF_WORD_256` theorem is bound in both the new files, but it's a harmless duplication and can also potentially be cleaned up in the followup refactoring.

PR #389, opened 2026-04-08
Add NIST SP 800-38D GCM specification (AES-128/256)

sgmenda:
This PR adds a high-level AES-GCM specification following [NIST SP 800-38D](https://doi.org/10.6028/NIST.SP.800-38D) and connects it to the existing NIST GHASH algebraic layer (#396), completing the chain from the spec down to `ghash_polyval_acc`.

The spec is parameterized over the block cipher, so the same definitions serve both AES-128 and AES-256.

### New files

**`common/gcm.ml`** — Standalone GCM specification:
- `inc32` — increment rightmost 32 bits of a 128-bit block (Section 6.2)
- `gctr` — counter-mode encryption (Algorithm 3), parameterized over the block cipher `enc`
- `gcm_ae` — authenticated encryption (Algorithm 4), parameterized over `enc`
- `gcm_ad` — authenticated decryption (Algorithm 5), parameterized over `enc`

The GF(2^128) multiply and GHASH fold reuse `nist_dot` and `nist_ghash` from `common/ghash_nist_defs.ml` rather than redefining them, so there is a single definition of each in `common/`.

Simplified to 96-bit IVs, full 128-bit blocks, 128-bit tags, and pre-expanded key schedules. The block cipher (`aes128_cipher` / `aes256_cipher`) comes from `common/fips197.ml`.

**`common/ghash_nist_defs.ml`** — Lightweight definitions extracted from `ghash_nist_bridge.ml`:
- `bit_reflect128`, `nist_dot`, `nist_ghash` definitions + simple lemmas
- Loadable without the heavy polynomial algebra (~1s vs minutes)

### Modified files

**`common/ghash_nist_bridge.ml`** — Refactored to `needs "common/ghash_nist_defs.ml"` instead of defining `bit_reflect128`, `nist_dot`, `nist_ghash` inline. Heavy proofs (`GUERON_PROP1`, `NIST_GHASH_IS_POLYVAL`, etc.) unchanged.

**`common/fips197.ml`** — Added `NIST_TC16_KEY_SCHEDULE`, the pre-expanded AES-256 key schedule for the McGrew-Viega TC16 key, alongside the existing AES-256 schedules. Validated by `aes256_cipher` proofs against known outputs (`E(K,J0)` and `H = E(K,0)`).

### Testing

KATs validated against [GCM revised spec test vectors](https://csrc.nist.rip/groups/ST/toolkit/BCM/documents/proposedmodes/gcm/gcm-revised-spec.pdf#section.B) (McGrew & Viega, Appendix B):

| Test case | Cipher | Plaintext | AAD | What it exercises |
|-----------|--------|-----------|-----|-------------------|
| TC1 | AES-128 | empty | empty | Bare GHASH + tag |
| TC2 | AES-128 | 1 zero block | empty | GCTR with plaintext |
| TC3 | AES-128 | 4 blocks | empty | Full pipeline, non-trivial key/IV |
| TC13 | AES-256 | empty | empty | AES-256 bare GHASH + tag |
| TC14 | AES-256 | 1 zero block | empty | AES-256 GCTR with plaintext |
| TC16 | AES-256 | 4 blocks | empty | AES-256 full pipeline, non-trivial key/IV |

The trivial `inc32` KATs run on load; the slower KATs are commented out for fast loading but can be run interactively. Each KAT is a deconstructed proof: individual steps (AES, GCTR, GHASH, tag XOR) are proved as separate theorems. The tag KATs derive `tag = S XOR E(K,J0)` from the computed GHASH and E(K,J0) theorems rather than re-pasted literals, so they cannot drift from the values the spec actually computes.

All definitions verified in a clean HOL Light session.

### How it connects

```mermaid
flowchart TD
gcm.ml-->|nist_ghash|ghash_nist_defs.ml-->|NIST_GHASH_IS_POLYVAL|ghash_nist_bridge.ml-->polyval_ghash.ml-->asm
```

`gcm.ml` uses `nist_ghash` directly, so no bridge theorem is needed between them. `nist_ghash` is defined in `ghash_nist_defs.ml`; `NIST_GHASH_IS_POLYVAL` (in `ghash_nist_bridge.ml`) bridges it to `ghash_polyval_acc`.


common/gcm.ml
```diff
@@ -0,0 +1,398 @@
...
+  let J0 : 128 word = word_join iv (word 1 : 32 word) in
+  let P = gctr ks (inc32 J0) C in
+  let len_block : 128 word =
+    word_join (word (128 * LENGTH A) : 64 word)
+              (word (128 * LENGTH C) : 64 word) in
+  let S = ghash H (word 0) (APPEND A (APPEND C [len_block])) in
+  let tag' = word_xor S (aes128_cipher J0 ks) in
+  if tag' = tag then SOME P else NONE`;;
+
+
```
sgmenda, line 272: @nebeid the the current KATs

common/gcm_nist_bridge.ml
nebeid, line 1: `bitrev128`, `gf128_mul` and `ghash` in `common/gcm.ml` are the same definitions as `bit_reflect128`, `nist_dot` and `nist_ghash` from #396. `int128` is a type abbreviation for `(128)word` (`Library/words.ml:437`), so after expansion they're identical up to bound-variable naming:

| `common/gcm.ml` (this PR) | merged (#396) |
|---|---|
| `bitrev128 x = word_reversefields 1 x` | `bit_reflect128 a = word_reversefields 1 a` |
| `gf128_mul X Y = bitrev128(ghash_reduce(word_pmul (bitrev128 X) (bitrev128 Y)))` | `nist_dot a b = bit_reflect128(ghash_reduce(word_pmul (bit_reflect128 a) (bit_reflect128 b)))` |
| `ghash H Y (CONS x rest) = ghash H (gf128_mul (word_xor Y x) H) rest` | `nist_ghash h acc (CONS x xs) = nist_ghash h (nist_dot (word_xor acc x) h) xs` |

Same operand order, same `ghash_reduce`, same `word_pmul`, same base case. That's why `BITREV128_IS_BIT_REFLECT128` and `GF128_MUL_IS_NIST_DOT` close by `REWRITE_TAC` on the definitions alone, and why `GHASH_IS_NIST_GHASH` needs induction only because they're two separate recursive constants rather than one.

To be clear about how this happened: it's a merge-order artifact, not anyone'sdesign decision. `common/gcm.ml` was drafted here in April on an unmerged branch; #396 landed in May and `common/gcm.ml` wasn't in the tree at that point, so neither side had anything to reuse. Two vocabularies grew in parallel. But now that they're about to coexist in `common/`, one set should go.

Concretely, what I'd suggest is the cheap direction — it makes this PR *smaller*:

- `common/gcm.ml` adds `needs "common/ghash_nist_defs.ml"` and drops its three definitions.
- `common/gcm_nist_bridge.ml` is then unnecessary and can go. Nothing `needs` it today, so it costs this PR nothing to drop.

That works cleanly because `gcm.ml` already does `needs "common/ghash.ml"` — which is where `ghash_reduce` itself lives — so the substantive algebra is already shared between the two files; only the thin wrappers were duplicated.

On the names, I want to separate that question out, because I think the naming preference actually points *your* way and I don't want the two to get conflated. `bitrev128` is closer to existing repo vocabulary than `bit_reflect128` is —`common/mlkem_mldsa.ml` already has `bitreverse7`/`bitreverse8` on the same `word_reversefields 1` idiom — and `gf128_mul` reads better than `nist_dot`. I'm happy for those names to stay.

What I'd ask is that the rename be its own commit rather than arriving as duplicate-plus-bridge, because the cost is very unevenly spread across the three:

- `bit_reflect128` → `bitrev128` is free. Neither it nor its four derived lemmas is referenced outside `ghash_nist_bridge.ml` anywhere.
- `nist_dot` → `gf128_mul` is free for the constant, but `NIST_DOT_IS_POLYVAL_DOT`  is named after it and *is* used — 27 of @jargh's kernel files, plus the AES-256-GCM encrypt proof. Renaming one without the other would be incoherent, so it's really  ~28 files. Maybe we can keep the lemma that proves them the same but move it to `ghash_nist_defs.ml`.
- `nist_ghash` is the expensive one: ~450 references across ~31 files on the three in-flight AES-GCM proof branches (encrypt, decrypt, and @jargh's kernel experiments), plus `NIST_GHASH_IS_POLYVAL` / `_APPEND` / `_CONS` / `_NIL`.

None of that is an argument against the rename — just an argument for doing it deliberately and in coordination, rather than as a side effect of landing a spec. The one name I'd push back on slightly is bare `ghash`: HOL Light's constant namespace is flat and `common/ghash.ml` already owns
`ghash_poly`/`mod_ghash`/`ghash_reduce`, so `nist_ghash` does carry useful information about *which* fold it is.

Either way, the thing I'd like to avoid is two names for one function in `common/`. They can silently diverge — revise the bit convention on one and the other doesn't follow, with only a lemma someone must remember to re-prove as coupling — and every downstream proof then has to pick a vocabulary and pay rewrite steps to cross.

nebeid (commented):
Thank you, Sanketh, for the revised PR. 

#### Request: AES-256 before this merges

As written, the spec is AES-128 only, and the hardcoding goes one level deeper than `gcm_ae`/`gcm_ad` — `gctr` has it too:

```ocaml
gctr:    CONS (word_xor x (aes128_cipher icb ks)) (gctr ks (inc32 icb) rest)
gcm_ae:  let H = aes128_cipher (word 0) ks in ... word_xor S (aes128_cipher J0 ks)
gcm_ad:  let H = aes128_cipher (word 0) ks in ... word_xor S (aes128_cipher J0 ks)
```

`inc32` and the GF(2^128)/GHASH layer are already cipher-agnostic, and `common/fips197.ml` already defines `aes256_cipher` — so the material is all here.

This matters because the AES-GCM kernels currently being proved against assembly are **AES-256**, so in its present form this spec can't serve as their correctness target.

I'd suggest doing it in this PR rather than as a follow-up, for two concrete reasons:

1. Once these definitions are in `common/`, changing `gctr`'s or `gcm_ae`'s signature becomes a breaking change to a public definition. Right now this PR has no dependents.
2. The KATs sit directly on `gctr` (`GCTR_CONV`, `GCTR_FAST_CONV`, `gctr_kat_1..3`) and on `gcm_ae`/`gcm_ad` via TC1/TC2/TC3. If the shape changes later, all of those need restating — better once, by whoever owns them, than twice by someone else afterwards.

Either shape works from a consumer's point of view: parameterise the existing definitions over the block cipher (cleaner end state), or add `gctr_256`/`gcm_ae_256`/`gcm_ad_256` alongside (some duplication, but leaves the existing KATs untouched). Happy with whichever you prefer.

Worth noting the composition point this enables: `gcm_ae` fixes `J0 = iv || 1` and starts from `gctr ks (inc32 J0)`, i.e. counter 2. The machine-code theorems are general in the starting counter, so they compose with this spec at that instance — but only against an AES-256 `gcm_ae`.

#### AES round/cipher equivalence

Agreed that `AES128_CIPHER_RECONSTRUCT` in [@jargh's branch](https://github.com/jargh/s2n-bignum-dev/blob/gcm/arm/proofs/aes_gcm_enc_kernel_x4_basic.ml) is the right shape — it proves the `aese`/`aesmc` chain equal to `word_reversefields 8 (aes128_cipher (word_reversefields 8 pt) (MAP ... rk))`, which subsumes the round-level step. Two observations:

- It's AES-128, so it doesn't yet cover the AES-256 kernels.
- Its ~45-line helper preamble (`AES_SUB_BYTES_SHIFT_ROWS`, `WORD_XOR_REVERSEFIELDS`, `AES_SUB_BYTES_REVERSEFIELDS`, `FIPS197_EQ_SHIFT_ROWS`, `FIPS197_EQ_MIX_COLUMNS`) depends only on
 `common/aes.ml`, `common/fips197.ml`, `arm/proofs/utils/aes_encrypt_spec.ml` and HOL Light's `Library/words.ml` — all already on `main`. So an AES-256 analogue is unblocked today and independent of this PR.

I'm happy to land that as a small standalone PR, including giving those five helper lemmas a shared home under `arm/proofs/utils/` rather than the copies currently duplicated across the kernel experiment files. @jargh — would you
rather own it, since it's your proof script ported from 10 rounds to 14?

sgmenda:
Thanks for the detailed review. I've pushed an update addressing the AES-256 request.

**Parameterized over the block cipher.** `gctr`, `gcm_ae`, and `gcm_ad` now take the block cipher `enc` as a parameter (rather than hardcoding `aes128_cipher`), threaded through the GCTR conversions. Instantiating with `aes128_cipher` or `aes256_cipher` from `common/fips197.ml` both work, so the same definitions serve AES-128 and AES-256. This is the "parameterise the existing definitions" shape you described as the cleaner end state, done here rather than as a follow-up per your point about avoiding a later breaking change to public definitions.

**AES-256 KATs.** Added TC13 and TC14 (AES-256 analogues of TC1/TC2) alongside the existing AES-128 cases, all validated against the NIST SP 800-38D / McGrew-Viega vectors.

**Deduplication.** While in here, `gcm.ml` now reuses `bit_reflect128`/`nist_dot`/`nist_ghash` from `common/ghash_nist_defs.ml` instead of defining its own `bitrev128`/`gf128_mul`/`ghash`, so there's a single definition of each in `common/`. That also let me drop `common/gcm_nist_bridge.ml`, which only existed to bridge those now-removed duplicates — `gcm.ml` uses `nist_ghash` directly, so the connection to the algebraic layer is definitional. The updated PR description has the revised file list and dependency chain.

**KAT hardening.** The tag KATs now derive `tag = S XOR E(K,J0)` from the computed GHASH and E(K,J0) theorems instead of re-pasted literals. This caught a stale literal in `tc2_tag` (it used the wrong GHASH value and produced the wrong tag while the comment claimed the correct one); with the values now derived from the computation, that class of drift can't recur.

On the **AES round/cipher equivalence** part: agreed it's independent of this PR and unblocked today. Happy to leave the AES-256 analogue of `AES128_CIPHER_RECONSTRUCT` (and giving the five helper lemmas a shared home under `arm/proofs/utils/`) as a separate PR — @jargh, over to you on whether you'd like to own that port since it's your proof script.

common/fips197.ml
```diff
@@ -665,3 +665,42 @@ prove(`aes256_cipher (word 0x014730f80ac625fe84f026c60bfd547d)
        word 0x5c9d844ed46f9885085e5d6a4f94c7d7`,
   CONV_TAC(LAND_CONV(FIPS197_ENCRYPT_FAST_CONV aes256_cipher
     AESAVS_ZERO_KEY_256_SCHEDULE)) THEN REFL_TAC);;
+
+(* ========================================================================= *)
+(* KAT: AES-256 with the McGrew-Viega GCM Test Case 16 key                   *)
```
nebeid: This is the **Test Case 15** key/vector, not Test Case 16.

The McGrew-Viega AES-128 -> AES-256 pairing runs at a **+12 offset**: TC1<->TC13, TC2<->TC14, TC3<->**TC15**, TC4<->TC16. Since the new KAT is the AES-256 analogue of TC3 (4 full blocks, empty AAD), it's TC15; the numbering here is off by one.

Evidence from the values in this PR:

| | this PR | TC15 | TC16 |
|---|---|---|---|
| tag | `b094dac5d93471bdec1a502270e3cc6c` | same | `76fc6ece0f4e1768cddf8853bb2d551b` |
| P | 4 full blocks, ends `1aafd255` | 64 B | 60 B, ends `ba637b39` |
| A | empty | empty | 20 B, `feedfacedeadbeeffeedfacedeadbeefabaddad2` |
| `len_block` | `0x...0200` (512 bits C, 0 bits A) | matches | would be `0x...01E0` + `0xA0` |

TC16 is the analogue of TC4, i.e. the one *with* AAD and a partial final block — a different and more demanding case.

Good news: **the data is all correct.** TC15 and TC16 share the same key and IV, so the key schedule and both `aes256_cipher` KATs below are right as they stand. It's purely a labelling fix: `NIST_TC16_KEY_SCHEDULE` -> `NIST_TC15_KEY_SCHEDULE`, this comment, the `tc16_*` bindings and header in `common/gcm.ml`, the `c97341e9` commit message, and the KAT table in the PR description.

Worth fixing before merge rather than after, since `NIST_TC16_KEY_SCHEDULE` becomes a public name in `common/` on merge, and anyone later adding the real TC16 (the AAD case) would find the name taken by a vector that isn't it.

Minor, separate: this is GCM test data living in the AES file. Defensible, since the two `aes256_cipher` KATs immediately below consume it — just flagging in case you'd rather the schedule sat in `common/gcm.ml` next to its only other user.

common/gcm.ml
```diff
@@ -0,0 +1,514 @@
...
+(* S = 0x83de425c5edc5d498f382c441041ca92 *)
+
+let tc14_aes_j0 = FIPS197_ENCRYPT_FAST_CONV aes256_cipher AESAVS_ZERO_KEY_256_SCHEDULE
+  `aes256_cipher (word 1 : 128 word) AESAVS_ZERO_KEY_256_SCHEDULE`;;
+
+let tc14_tag = mk_tag tc14_ghash tc14_aes_j0;;
+(* T = 0xd0d1c8a799996bf0265b98b5d48ab919 *)
+
+(* ========================================================================= *)
+(* Test Case 16: AES-256, 4-block P, empty A                                 *)
```
nebeid: Same labelling point as on `common/fips197.ml`: this block is **Test Case 15**. The tag `b094dac5d93471bdec1a502270e3cc6c`, the four full ciphertext blocks, and `len_block = 0x...0200` (512 bits of C, 0 bits of A) all identify TC15; TC16 has 20 bytes of AAD, a 60-byte plaintext and tag `76fc6ece0f4e1768cddf8853bb2d551b`.

The header comment here already describes the TC15 shape ("4-block P, empty A", and "AES-256 analogue of TC3" — TC3's analogue *is* TC15), so only the number is wrong. Affects this comment plus the `tc16_*` binding names below.

common/gcm.ml
```diff
@@ -0,0 +1,514 @@
...
+(* GCTR/GCM-AE/GCM-AD take the block cipher as a parameter, so the same      *)
+(* definitions serve AES-128 and AES-256 (from common/fips197.ml).           *)
+(*                                                                           *)
+(* The GF(2^128) multiply and GHASH fold are reused from                     *)
+(* common/ghash_nist_defs.ml (nist_dot, nist_ghash) rather than redefined    *)
+(* here, so there is a single definition of each in common/.                 *)
+(* ========================================================================= *)
+
+needs "common/fips197.ml";;
+needs "common/ghash.ml";;
```
nebeid: Nit: this `needs` is now redundant — `common/ghash_nist_defs.ml` on the next line already does `needs "common/ghash.ml"`. Harmless either way, and there's a case for keeping it as documentation of the direct dependency on `ghash_reduce`; just noting it in case you'd rather drop the line.

common/fips197.ml
nebeid: This is the **Test Case 15** key/vector, not Test Case 16.

The McGrew-Viega AES-128 -> AES-256 pairing runs at a **+12 offset**: TC1<->TC13, TC2<->TC14, TC3<->**TC15**, TC4<->TC16. Since the new KAT is the AES-256 analogue of TC3 (4 full blocks, empty AAD), it's TC15; the numbering here is off by one.

Evidence from the values in this PR:

| | this PR | TC15 | TC16 |
|---|---|---|---|
| tag | `b094dac5d93471bdec1a502270e3cc6c` | same | `76fc6ece0f4e1768cddf8853bb2d551b` |
| P | 4 full blocks, ends `1aafd255` | 64 B | 60 B, ends `ba637b39` |
| A | empty | empty | 20 B, `feedfacedeadbeeffeedfacedeadbeefabaddad2` |
| `len_block` | `0x...0200` (512 bits C, 0 bits A) | matches | would be `0x...01E0` + `0xA0` |

TC16 is the analogue of TC4, i.e. the one *with* AAD and a partial final block — a different and more demanding case.

Good news: **the data is all correct.** TC15 and TC16 share the same key and IV, so the key schedule and both `aes256_cipher` KATs below are right as they stand. It's purely a labelling fix: `NIST_TC16_KEY_SCHEDULE` -> `NIST_TC15_KEY_SCHEDULE`, this comment, the `tc16_*` bindings and header in `common/gcm.ml`, the `c97341e9` commit message, and the KAT table in the PR description.

Worth fixing before merge rather than after, since `NIST_TC16_KEY_SCHEDULE` becomes a public name in `common/` on merge, and anyone later adding the real TC16 (the AAD case) would find the name taken by a vector that isn't it.

Minor, separate: this is GCM test data living in the AES file. Defensible, since the two `aes256_cipher` KATs immediately below consume it — just flagging in case you'd rather the schedule sat in `common/gcm.ml` next to its only other user.

common/gcm.ml
nebeid: Same labelling point as on `common/fips197.ml`: this block is **Test Case 15**. The tag `b094dac5d93471bdec1a502270e3cc6c`, the four full ciphertext blocks, and `len_block = 0x...0200` (512 bits of C, 0 bits of A) all identify TC15; TC16 has 20 bytes of AAD, a 60-byte plaintext and tag `76fc6ece0f4e1768cddf8853bb2d551b`.

The header comment here already describes the TC15 shape ("4-block P, empty A", and "AES-256 analogue of TC3" — TC3's analogue *is* TC15), so only the number is wrong. Affects this comment plus the `tc16_*` binding names below.

common/gcm.ml
nebeid: Nit: this `needs` is now redundant — `common/ghash_nist_defs.ml` on the next line already does `needs "common/ghash.ml"`. Harmless either way, and there's a case for keeping it as documentation of the direct dependency on `ghash_reduce`; just noting it in case you'd rather drop the line.

common/fips197.ml
nebeid: This is the **Test Case 15** key/vector, not Test Case 16.

The McGrew-Viega AES-128 -> AES-256 pairing runs at a **+12 offset**: TC1<->TC13, TC2<->TC14, TC3<->**TC15**, TC4<->TC16. Since the new KAT is the AES-256 analogue of TC3 (4 full blocks, empty AAD), it's TC15; the numbering here is off by one.

Evidence from the values in this PR:

| | this PR | TC15 | TC16 |
|---|---|---|---|
| tag | `b094dac5d93471bdec1a502270e3cc6c` | same | `76fc6ece0f4e1768cddf8853bb2d551b` |
| P | 4 full blocks, ends `1aafd255` | 64 B | 60 B, ends `ba637b39` |
| A | empty | empty | 20 B, `feedfacedeadbeeffeedfacedeadbeefabaddad2` |
| `len_block` | `0x...0200` (512 bits C, 0 bits A) | matches | would be `0x...01E0` + `0xA0` |

TC16 is the analogue of TC4, i.e. the one *with* AAD and a partial final block — a different and more demanding case.

Good news: **the data is all correct.** TC15 and TC16 share the same key and IV, so the key schedule and both `aes256_cipher` KATs below are right as they stand. It's purely a labelling fix: `NIST_TC16_KEY_SCHEDULE` -> `NIST_TC15_KEY_SCHEDULE`, this comment, the `tc16_*` bindings and header in `common/gcm.ml`, the `c97341e9` commit message, and the KAT table in the PR description.

common/gcm.ml
nebeid: Same labelling point as on `common/fips197.ml`: this block is **Test Case 15**. The tag `b094dac5d93471bdec1a502270e3cc6c`, the four full ciphertext blocks, and `len_block = 0x...0200` (512 bits of C, 0 bits of A) all identify TC15; TC16 has 20 bytes of AAD, a 60-byte plaintext and tag `76fc6ece0f4e1768cddf8853bb2d551b`.

The header comment here already describes the TC15 shape ("4-block P, empty A", and "AES-256 analogue of TC3" — TC3's analogue *is* TC15), so only the number is wrong. Affects this comment plus the `tc16_*` binding names below.

common/gcm.ml
nebeid: Nit: this `needs` is now redundant — `common/ghash_nist_defs.ml` on the next line already does `needs "common/ghash.ml"`.

common/gcm.ml
```diff
@@ -0,0 +1,514 @@
...
+(* This is nist_ghash from common/ghash_nist_defs.ml, reused here.           *)
+(* Conversions for stepping it on concrete blocks live here.                 *)
+(* ========================================================================= *)
+
+let NIST_GHASH_STEP_CONV =
+  ONCE_REWRITE_CONV [nist_ghash] THENC
+  DEPTH_CONV WORD_RED_CONV THENC
+  ONCE_DEPTH_CONV NIST_DOT_CONV;;
+
+let rec NIST_GHASH_CONV tm =
```
nebeid: `NIST_GHASH_CONV` is never called — its only two references in the tree are this definition and the self-recursive call on the next line. `GCTR_CONV` (:65) is the same shape, and `GCTR_STEP_CONV` (:56) is reachable only through it, so it's transitively unreachable too. Every KAT in the file goes via `NIST_GHASH_STEP_CONV` (22 references) or `GCTR_FAST_CONV` (9) instead.

This isn't something the update introduced — `GHASH_CONV` and `GCTR_CONV` were unused the same way in the previous revision. Raising it because the file becomes public API in `common/` on merge, and the concern isn't really that they're unused: it's that they're **unvalidated**. If a recursive driver is wrong, nothing here would show it, since the KATs exercise the other variants.

There's a fix that costs nothing and improves the KATs: the TC3 and TC15 GHASH folds currently step five times by hand (`tc3_gh0..gh4`, `tc16_gh0..gh4`), which is exactly what `NIST_GHASH_CONV` automates. Using the wrapper for one of them would exercise it *and* shorten the KAT. I'd keep the explicit per-step form for the other — stepwise really is easier to debug when a vector disagrees — so this is about having each variant covered once, not converting them all.

Same idea for `GCTR_CONV`: one KAT through the slow path alongside the `GCTR_FAST_CONV` ones. Or, if the slow variants aren't earning their place, deleting `GCTR_CONV`/`GCTR_STEP_CONV` and keeping only the FAST pair is an equally good outcome — either way nothing exported goes untested.

common/fips197.ml
sgmenda: You're right, it's TC15. Fixed.

common/gcm.ml
sgmenda: Good point. Added KAT cross-checking. Did the same for GCTR_CONV vs GCTR_FAST_CONV. All commented out since they are rather slow.

nebeid: Thank you, Sanketh, for the addition. Did you consider using them in some of the KATs as suggested?

sgmenda: Good call. Addressed in c07be713e4d5c86d9ca06d86c440ed07bb03e6d5

nebeid (approved):

jargh (approved):
All looks great to me, and Nevine's earlier input has clearly improved things. Thanks for systematically getting this into shape! I'll approve and merge.

PR #391, opened 2026-04-16
Use HOL Light's new record library

aqjune-aws:
This is a simple patch that makes s2n-bignum use the record Library recently migrated to HOL Light. :)

Patch written by Claude


jargh (approved):
Great, thank you, it clearly makes sense to take this from upstream now. I will merge.

PR #397, opened 2026-04-28
x86: Add VPCMPGTD, VPCMPGTW instruction models

jakemas:
## Summary
- Add VEX-encoded `VPCMPGTD` (packed signed dword compare greater-than) and `VPCMPGTW` (packed signed word compare greater-than) instruction models
- Both 128-bit (xmm) and 256-bit (ymm) variants supported
- Needed for the ML-DSA `caddq` (conditional add Q) formal verification

## Changes
- `x86/proofs/instruction.ml`: Add `VPCMPGTD` and `VPCMPGTW` to instruction type
- `x86/proofs/decode.ml`: Add VEX decoder entries for opcodes `0x66` (VPCMPGTD) and `0x65` (VPCMPGTW)
- `x86/proofs/x86.ml`: Add semantic definitions, execution dispatch, ALT expansion rules, and operation clauses
- `x86/proofs/simulator.ml`: Add 32 cosimulation test cases (ymm/xmm, low/high registers)
- `x86/allowed_asm`: Add `vpcmpgtd` and `vpcmpgtw`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #398, opened 2026-04-28
ML-DSA x86 HOL-Light proof mldsa_caddq

jakemas:
## Summary
- Add x86 AVX2 implementation of `mldsa_caddq` (conditional add Q for ML-DSA polynomial coefficients)
- Complete formal verification with 12 theorems: 4 functional correctness + 4 memory safety + 4 internal lemmas
- Proves: for inputs with `|ival(x)| < Q`, the output satisfies `ival(output) = ival(x) rem Q`

**Depends on:** #397 (VPCMPGTD/VPCMPGTW instruction models)

## Changes
- `x86/mldsa/mldsa_caddq.S`: Fully unrolled AVX2 assembly (VPCMPGTD + VPAND + VPADDD)
- `x86_att/mldsa/mldsa_caddq.S`: AT&T syntax version (auto-generated, binary-verified)
- `x86/proofs/mldsa_caddq.ml`: Complete formal proof (815 lines)
- `x86/proofs/specifications.txt`: 8 theorem entries (4 correct + 4 safe)
- `x86/proofs/subroutine_signatures.ml`: Signature for safety proof
- `x86/Makefile`, `x86_att/Makefile`: Build entries
- `include/s2n-bignum.h`, `include/s2n-bignum-c89.h`: Function prototype
- `tests/test.c`: Functional test with reference implementation
- `benchmarks/benchmark.c`: Benchmark entry
- `tools/collect-signatures.py`: Added to onlyInX86

## Test plan
- [x] Proof verified in HOL Light (clean `loadt` from checkpoint)
- [x] Safety proofs verified (all 4 SAFE variants)
- [x] AT&T syntax binary-matches Intel syntax (`make` in `x86_att/`)
- [x] `cd tests && make go` — mldsa_caddq passes 100 test cases
- [x] `cd benchmarks && make go` — 14.3 ns/call
- [x] Cross-platform build (x86 + macOS)
- [x] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jargh:
All looks good, though here are a few suggestions for improvements:
*  There's a duplication in `x86/allowed_asm` of `vpcmpgtd` and `vpcmpgtw`, probably just a merge thing (and the new ones are out of alphabetical order).
* The code saves and restores `xmm6` for Windows but I think it's actually not needed. Doesn't really matter but if it's not too disruptive, might as well clean up
* The correctness specs are nice clean mathematics but a bit redundant in that the `result = val(x i) rem 8380417` conjunct implies easily (via `INT_REM_POS` and `INT_LT_REM`) the other conjunct, i.e. that `0 <= result < 8380417`. On the one hand it does no harm to call it out explicitly but on the other it makes the spec a bit more cluttered/redundant. I don't feel that strongly but there could be an argument for just skipping the second conjunct of each postcondition.

jakemas:
Thanks John! Addressed the first two:

- **`allowed_asm` duplication**: removed the duplicate `vpcmpgtd`/`vpcmpgtw` entries and they're back in alphabetical order.
- **`xmm6` save/restore**: good catch — the body only uses ymm0–ymm5 (all volatile in the Windows ABI), so the xmm6 spill was dead. Simplified the prologue/epilogue to a plain `push rdi` / `pop rdi` and updated the Windows HOL Light proofs to match.

On the third (the redundant `0 <= result < 8380417` conjunct): you're right that it follows from `result = ival(x i) rem 8380417` via `INT_REM_POS`/`INT_LT_REM`. We've actually been keeping it deliberately — it mirrors the CBMC contract on the mldsa-native side, where `mld_poly_caddq_avx2_asm`'s `ensures(array_bound(...))` is exactly this bound and is marked "must be kept in sync with the HOL-Light specification". That's tracking towards pq-code-package/mldsa-native#673 (prove the backend CBMC contract implies the generic frontend contract), so stating the bound explicitly keeps the two specs in lock-step. That said, I don't feel strongly either — happy to drop the second conjunct for the s2n-bignum copy if you'd prefer the cleaner mathematics. Let me know and I'll strip it.

jargh:
OK, understood - that makes good sense, thanks!

jargh (approved):
This is now all great, thanks! I'll merge.

PR #400, opened 2026-05-05
BIF, REV32, and MOVI Instructions

manastasova:
Instruction changes:                                                   
  - **BIF (Bitwise Insert if False)** - NEON vector insert, complement of the existing `BIT`. Operates on 64- and 128-bit vectors.
  - **REV32 (vector)** — byte-reverse within each 32-bit element.
  - **MOVI (8-bit replicated byte immediate)** - replicates an 8-bit immediate into every byte of the destination.

### DIFF
--- a/arm/proofs/decode.ml
+++ b/arm/proofs/decode.ml
@@ -105,7 +105,10 @@ let arm_ldst3 = new_definition `arm_ldst3 ld Rt1 =
 let arm_adv_simd_expand_imm = new_definition
   `(arm_adv_simd_expand_imm:(8)word->(1)word->(4)word->((64)word)option)
       abcdefgh op cmode =
-    if val cmode = 14 /\ val op = 1 then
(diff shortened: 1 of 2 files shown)


jakemas:
Check out https://github.com/awslabs/s2n-bignum/pull/378/changes#top as `MOVI` is also added in there

jargh:
Now that https://github.com/awslabs/s2n-bignum/pull/406 has been merged, a lot of this is subsumed (and you'll see that the content is very similar.) One option is to just close this PR but, if you like, it could be turned into one that adds the additional support for REV32 in the 64-bit case, which is not covered by the existing PR 406 merge. I'll leave the choice to you :-)

manastasova:
Closing as instructions are added in https://github.com/awslabs/s2n-bignum/pull/406. If REV32 in the 64-bit case is needed later on, it will be added via new PR.

PR #401, opened 2026-05-06
ML-DSA x86 AVX2 rej_uniform CORRECT and MEMSAFE proofs

jakemas:
*Issue #, if available:*

Adds CORRECT and MEMSAFE proofs of the x86_64 AVX2 `rej_uniform.S` used in
- mldsa-native: https://github.com/pq-code-package/mldsa-native/blob/main/mldsa/src/native/x86_64/src/rej_uniform_avx2.S
- dilithium (reference): https://github.com/pq-crystals/dilithium/blob/master/avx2/rejsample.S

This function implements ML-DSA's 23-bit rejection sampling. A main AVX2 loop processes 24 bytes (8 coefficients) per iteration via VPERMQ + VPSHUFB extraction, VPAND masking, VPSUBD + VMOVMSKPS rejection, and VPERMD + table compaction; a scalar tail handles any remaining bytes.

*Description of changes:*

Adds proof for

- `MLDSA_REJ_UNIFORM_CORRECT`
- `MLDSA_REJ_UNIFORM_MEMSAFE`

The specification matches #378's ARM counterpart: after the function the output buffer contains `num_of_wordlist (SUB_LIST (0, 256) (REJ_SAMPLE inlist))` with RAX = its length, where `REJ_SAMPLE` is the FIPS-204 reference filter


jargh:
Looks great in essentials, with a few minor nits about keeping things consistent with the usual style:
* There seem to be more saves/restores than needed in the Windows prologue/epilogue - as far as I can see we're using few if any of the higher ymm registers in the body. May be able to get away with something more like the mlkem analog.
* We're missing the S2N_BN_SIZE_DIRECTIVE macro at the end
* It'd be nice to use CFI-wrapped macros for the stack operations (admittedly they might be less useful on Windows anyway and this mainly affects the Windows wrappers)
* For consistency with the other functions, let's use the VARIABLE_TIME suffix in the filename as well as the code.

jargh (approved):
All looks great now, thanks! I'll merge.

PR #402, opened 2026-05-06
ML-DSA AArch64 HOL-Light CORRECT and MEMSAFE proofs for mldsa_rej_uniform_eta{2/4}

jakemas:
Porting proofs from mldsa-native https://github.com/pq-code-package/mldsa-native/pull/1040

## Summary

Add formal verification proofs for `mldsa_rej_uniform_eta2` and `mldsa_rej_uniform_eta4`, the AArch64 NEON rejection-sampling routines used in ML-DSA's secret-key sampling (eta=2 for parameter sets 44/87, eta=4 for parameter set 65). The proofs establish full functional correctness against the FIPS 204-aligned spec `REJ_SAMPLE_ETA{2,4}` operating on a `(4 word) list` of nibbles: for each accepted nibble `n` (the predicate is `n < 15` for eta=2 and `n < 9` for eta=4), the assembly produces the int32 mapping (`2 - (n mod 5)` for eta=2, `4 - n` for eta=4), and writes up to 256 results to the output buffer matching `num_of_wordlist(SUB_LIST(0,256) (REJ_SAMPLE_ETA{2,4} inlist))`. Both routines also have memory-safety proofs (`_SUBROUTINE_MEMSAFE`); constant-time is not claimed since rejection sampling depends on private data.

## CORRECT Proof architecture

- **WOP characterization** of loop iteration count N (smallest i where the buffer is exhausted or 256 samples already collected).
- **ENSURES_WHILE_UP_TAC** over the main 240-byte sampling loop: 75-step preamble + 16 unrolled body iterations + back-edge + post-loop exit.
- **Writeback phase** (245 ARM steps): SUB+SSHLL cascade unpacking nibbles to int32 with sign extension of (4 − n) for eta=4 (or `(2 - n MOD 5)` via Barrett reduction `sqdmulh by 6554` for eta=2), stored via 64-byte chunks to `res`. Split into Case A (niblen ≥ 256, uses full stack) and Case B (niblen < 256).
- **Case B small** (niblen < 256) uses a *virtual stack list* L of length 256 materialized via `BYTES_EXISTS_WORDLIST`: the TBL writeback leaves garbage past the committed niblist, so the 1024-byte output identity doesn't hold against a zero-padded `STACK_CONTENT`. Instead we define L so that `bytes(sp, 512) s = num_of_wordlist L` (pure existence over bytes), prove `niblist = SUB_LIST(0, niblen) L`, run the Case A closure chain on L, and MOD-truncate the 1024-byte identity to `4*niblen`.

The `table` argument is a 256-entry (16 bytes each) TBL permutation lookup indexed by 8-bit acceptance masks. The routines use 576 bytes of stack as a scratch area for nibble packing.


arm/proofs/mldsa_rej_uniform_eta2.ml
```diff
@@ -0,0 +1,3190 @@
...
+(* Constant-time and memory safety proof.                                    *)
+(*                                                                           *)
+(* Strategy: follow mlkem-native's mlkem_rej_uniform_VARIABLE_TIME.ml MEMSAFE *)
+(* pattern. PROVE_SAFETY_SPEC_TAC's auto-symbolic-execution fails for variable*)
+(* loops with pure-register sequences > chunk size (50). Instead, we write a *)
+(* custom MEMSAFE spec with ENSURES_SEQUENCE_TAC + ENSURES_WHILE_UP_TAC and  *)
+(* event-tracking loop invariants, then DISCHARGE_MEMSAFE_TAC at boundaries. *)
+(* ========================================================================= *)
+
+needs "s2n_bignum/arm/proofs/consttime.ml";;
```
dkostic: wrong path, same in the other proof file.

jakemas: aha yup, thanks! Fixed now!

jargh (approved):
All looks good, thank you!

Issue #407, opened 2026-05-21
CI Improvement when running proofs


jakemas:
Currently, CI takes about ~3 hours to complete the ARM/X86 s2n-bignum proofs. This can be problematic when attempting to use the CI to validate that proofs are passing, or when iterating on PRs.

mldsa/mlkem-native splits up the proofs to:
- only run in CI when changes are made to proof files/dependent files (this could save us CI time/costs -- i.e. don't run the ed25519 proofs when I make changes to ml-dsa ones). It's still a good idea to run all dependent proofs on any changed common files, e.g. any change to common/mlkem_mldsa.ml would trigger all proofs that use it to re-run.
- split up each proof as its own CI job, that way the result of running short proof, say 10min total running time, would appear after 10min, and not after the longest proof in the batch finishes (currently 3h).

It would be lovely to make those same changes here

PR #408, opened 2026-05-22
[TEST] CI: split s2n-bignum-proofs into per-subdirectory matrix entries

jakemas:
## Summary
Replace the single arch-level proofs job (one ~3h run for all 341 arm proofs and 347 x86 proofs combined) with a per-subdirectory matrix that groups proofs by their assembly subdirectory. Each matrix entry runs only \`make subdir/*.correct\` for its subdirectory.

Failures now surface per-subdirectory on the PR page, and the critical-path wall-clock drops to the longest single subdirectory (typically \`generic/\`) instead of being summed across all of them.

Towards #407.

## Caveats / things to verify
- This relies on the CodeBuild fleet being able to spin up multiple instances per workflow run. The current \`runs-on\` qualifier \`-\${{ github.run_id }}-\${{ github.run_attempt }}\` may serialize matrix entries onto a single ephemeral instance — if that's the case, we'll see worse wall-clock than the status quo (per-job HOL-Light setup multiplied by 23 entries) and need to revisit the runner topology.
- Tutorial removed from this job (already covered by \`s2n-bignum-tutorial\` and \`s2n-bignum-tutorial-macos-arm\`).
- \`collect-times.sh\` removed; can be re-added later as an aggregator step.

## Test plan
- [ ] Watch this PR's CI to see whether the 23 matrix entries run in parallel or serialize
- [ ] Compare total wall-clock to the current 3h baseline
- [ ] Confirm a deliberate proof failure surfaces in only the relevant matrix entry


jakemas:
Testing some stuff out. A algorithm split is very simple to do: https://github.com/awslabs/s2n-bignum/pull/408/checks?sha=b1b785b3cafb16e7e78a088f6ac3306889f5bbc5. But I like the individual break down of each proof. Problem is, ~600+ proofs here total, so it's a balance of not overwhelming github runners. 

Current approach, matrix all 600+ proofs out, turn off all proofs not effectedm by changes. Instruction additions will be hefty, i suppose 300 CI jobs, so perhaps if the count gets above a particular theshold we default back to the non-granular versions that run today, and test everything in 3h

PR #409, opened 2026-06-02
x86: Add VPSUBB VPMOVMSKB VPMOVSXBD VPMOVZXBW instructions for ML-DSA eta4/eta2 rejection sampling

jakemas:
## Summary

Add support for 4 AVX2 instructions required for efficient ML-DSA eta4/eta2 rejection sampling:

- **VPSUBB** (VEX.0F.F8) - Packed byte subtraction
- **VPMOVMSKB** (VEX.0F.D7) - Create mask from most significant bit of each byte
- **VPMOVSXBD** (VEX.0F38.21) - Sign-extend 8 packed 8-bit integers to 32-bit integers
- **VPMOVZXBW** (VEX.0F38.30) - Zero-extend 16 packed 8-bit integers to 16-bit integers

These instructions are used in the AVX2-optimized rejection sampling routines for ML-DSA parameter sets with eta=4 and eta=2.

## Changes

Modified 4 files totaling 184 lines:

1. **x86/proofs/instruction.ml** - Added 4 instruction type constructors in alphabetical order
2. **x86/proofs/x86.ml** - Added execution semantics (4 main definitions + 4 _ALT versions)
3. **x86/proofs/decode.ml** - Added VEX decode patterns for all 4 instructions
4. **x86/proofs/simulator.ml** - Added 40 test cases covering register and memory operands


jargh:
Looks good for register operands but for memory operands this can read too large a chunk (16 versus 8 bytes). It's true that we're currently not using memory operands in any code, but if we did, this overread would at least matter for memory safety proofs. The PR (over)simplifies the existing decode of `VPMOVZXBD` in the same style -- I think in fact the deleted code should be how all three of them decode, so that they correctly handle memory operands. It's a bit verbose so a refactor is fine, but I think some such more complex decode _is_ truly needed.

jakemas:
> Looks good for register operands but for memory operands this can read too large a chunk (16 versus 8 bytes). It's true that we're currently not using memory operands in any code, but if we did, this overread would at least matter for memory safety proofs. The PR (over)simplifies the existing decode of `VPMOVZXBD` in the same style -- I think in fact the deleted code should be how all three of them decode, so that they correctly handle memory operands. It's a bit verbose so a refactor is fine, but I think some such more complex decode _is_ truly needed.

thank you, let me know what you think of [8866086](https://github.com/awslabs/s2n-bignum/pull/409/commits/8866086fd01301d2270bd3b15da160687f910412)

jargh:
All looks great functionally to me now and passes all tests, thanks! A couple of suggestions, enacted in the attached patch file:
* Aesthetically, it'd be nicer to use a more compact (e.g. summation-based) definition for `x86_VPMOVMSKB` and prove the explicit expansion only in the `x86_VPMOVMSKB_ALT` form -- that makes the actual specs more human-surveyable, which is an important thing to strive for.
* It might be worth adding a few trivial memory cases to the simulator just to sanity-check those last changes (I've run it myself locally and it seems fine). We aren't fully consistent about always doing this in general but since it was something that came up in the review here, not a bad idea.

Hope I haven't broken anything with this patch :-)
[pr409_refine.patch](https://github.com/user-attachments/files/28854400/pr409_refine.patch)

jargh:
It looks a bit funny that three successive instructions will then use different styles to express a similar semantics: `x86_PMOVMSKB` uses a clever `usimd16` trick, `x86_VMOVMSKPS` uses an explicit expansion and `x86_VPMOVMSKB` uses a sum. But at least each one is reasonably short and digestible :-)

jargh (approved):
Looks great now - the proofs have also mostly completed successfully and I'm sure nothing will break in specific proofs as a result of the latest change. I'll merge :-)

PR #410, opened 2026-06-04
x86: Add VPABSD and VPTEST instruction models

jakemas:
## Summary

Add semantic models for two AVX2 instructions required by mldsa-native's
`poly_chknorm` formal verification:

- **VPABSD** (VEX.66.0F38.WIG 1E `/r`) — packed signed absolute value of 32-bit
  integers. Used to take `|a[i]|` before comparing against a bound.
- **VPTEST** (VEX.66.0F38.WIG 17 `/r`) — bitwise AND test; sets `ZF` if AND is
  all zero, `CF` if `(NOT dest) AND src` is all zero, clears `AF/OF/PF/SF`.
  Used to detect whether any lane of an OR-accumulated mask is non-zero.

## Changes

- `x86/proofs/instruction.ml` -  2 new instruction constructors (alphabetical order) 
- `x86/proofs/x86.ml` - Semantic definitions + `_ALT` registrations + dispatch 
- `x86/proofs/decode.ml` - VEX decode patterns for 0F38 0x1E (VPABSD) and 0x17 (VPTEST), in 128-bit and 256-bit forms 
- `x86/proofs/simulator.ml` - 24 test cases covering register-to-register encodings with low/high register numbers 

## Notes

- Decoder restricts both opcodes to VEX.vvvv = 1111 per Intel SDM.
- VPABSD semantics use `usimd8` over a unary abs lane
  (`if word_igt 0 a then word_neg a else a`), matching the form produced
  by GCC for `_mm256_abs_epi32`.
- VPTEST semantics correctly clear flags AF/OF/PF/SF and set ZF/CF
  per the SDM truth table.


jargh (approved):
Thanks, it all looks flawless to me, and all tests and proofs confirm correctness. I'll merge.

PR #413, opened 2026-06-12
x86: Move edwards25519_scalarmulbase{,_alt} const tables to .rodata

aqjune-aws:
Same pattern as #242: place the precomputed tables in .rodata instead of .text and replace per-block rip-relative LEAs with a single LEA to the table base plus register-relative offsets. The corresponding HOL Light proofs are updated.

This is almost entirely written by Claude Code.

*Issue #, if available:*

*Description of changes:*


jargh (approved):
This is all excellent, thank you! It's great now to have achieved the milestone of moving all constant data into `.rodata` everywhere :-)

PR #414, opened 2026-06-20
x86: Move two missing edwards25519_scalarmuldouble{,_alt} const tables to .rodata

aqjune-aws:
Same pattern as #413: place the precomputed table in .rodata instead of .text and replace the per-block rip-relative LEAs with a LEA to the table base. The HOL Light proofs are updated to load the code via define_assert_relocs_from_elf (parametrized over pc and the table address `tables`), add nonoverlapping conditions between the constant table and the output buffers, and fold the rip-relative LEA results via RIP_REL_ADDR_FOLD under riprel32_within_bounds preconditions.

This patch was entirely written by Claude Code.

*Issue #, if available:*

*Description of changes:*


jargh (approved):
Excellent, thank you! Now we've *really* completed the `rodata` switch :-)

PR #416, opened 2026-06-23
x86: Adds an 8-bit instance of the CMP instruction model

torben-hansen:
*Description of changes:*

X86_OPERATION_CLAUSES previously carried x86_CMP only at :64 (x86_CMP_ALT)
and :32. Byte-sized memory/immediate compares dispatch to x86_CMP at :8,
which had no matching rewrite clause, so the symbolic stepper left the term
unexpanded and the step failed (surfacing as the generic "Equality between
two states is ill-formed" error).

simulator.ml gains the byte-CMP encoding (cmp BYTE PTR [rsp], 0) in its
extra-checks list so sematest cosimulates the new clause against hardware.

Needed to simplify https://github.com/awslabs/s2n-bignum/pull/415 proof (which created it's own workaround).


jargh (approved):
Clearly sound semantically. I've tried a toy example using the newly modeled instruction and the behavior in simulation was just what I would expect/hope. I'll merge.

PR #418, opened 2026-06-30
x86 Keccak safety proofs

manastasova:
# x86: Add constant-time and memory-safety proofs for SHA3 Keccak-f1600

Adds the full 5-theorem `*_SAFE` chains (`_SAFE`, `_NOIBT_SUBROUTINE_SAFE`, `_SUBROUTINE_SAFE`, `_NOIBT_WINDOWS_SUBROUTINE_SAFE`, `_WINDOWS_SUBROUTINE_SAFE`) for the three x86 Keccak-f1600 implementations:

- `sha3_keccak_f1600` — scalar
- `sha3_keccak4_f1600` — AVX2, 4-way batched
- `sha3_keccak4_f1600_alt` — AVX2, 4-way batched, using the rho8/rho56 rotate tables

## AVX2 subroutine lifts are not one-liners

- The scalar `sha3_keccak_f1600` lifts are via the trivial one-liners `X86_PROMOTE_RETURN_STACK_TAC` / `WINDOWS_X86_WRAP_STACK_TAC`.
- The two AVX2 functions dynamically realign `rsp` to a 32-byte boundary (`and rsp, …e0`) before spilling vector registers.
- The stock promote/wrap tactics assume a fixed frame and cannot bridge the caller's `stackpointer` to the realigned one, so the AVX2 subroutine theorems are lifted manually from the core `_SAFE`, mirroring the `delta` / `word_and` reasoning already in the matching correctness proofs.
- The Windows AVX2 lifts additionally reuse the `MAYCHANGE_*_QUARTER` XMM6–15 callee-save rewrites from the Windows correctness proofs.

## Why `tools/collect-signatures.py` changes

- `sha3_keccak4_f1600_alt` is the only function with a per-architecture signature: x86 (AVX2) takes two extra read-only args `rho8`/`rho56`; the generic/AArch64 variant does not (`#ifdef __x86_64__` in `include/s2n-bignum.h`).
- The safety proof needs `rho8`/`rho56` listed as input buffers so `mk_safety_spec` treats them as public args (like `rc`) — which means adding them to the shared `// Inputs …; output …` header comment the generator reads.
- That comment is shared across arches, but buffer element sizes are looked up from each arch's own arg list — so on AArch64 (no `rho8`/`rho56`) the generator would `KeyError`.
- Fix: skip emitting any buffer whose argument is absent from the architecture being generated — semantically correct (a buffer only exists for an existing argument) and a verified no-op for every function except `sha3_keccak4_f1600_alt`.
- `include/s2n-bignum-c89.h` and both `subroutine_signatures.ml` tables are regenerated accordingly.

## Testing

- `make proofs` builds all three keccak proofs green (each `.correct` ends with a `Running time:` line, prints its 4 `_SUBROUTINE_SAFE` theorems, no `error:`/`exception:`).
- Regenerating signatures changes only the intended x86 rho8/rho56 buffer lines, nothing else.


x86/proofs/sha3_keccak4_f1600.ml
```diff
@@ -1353,3 +1353,381 @@ let SHA3_KECCAK4_F1600_WINDOWS_SUBROUTINE_CORRECT = prove
...
+           ==> ensures x86
+               (\s.
+                    bytes_loaded s (word pc) sha3_keccak4_f1600_tmc /\
+                    read RIP s = word pc /\
+                    read RSP s = stackpointer /\
+                    read (memory :> bytes64 stackpointer) s = returnaddress /\
+                    C_ARGUMENTS [bitstate_in; rc_pointer] s /\
+                    read events s = e)
+               (\s.
+                    read RIP s = returnaddress /\
```
aqjune-aws, line 1440: 🟡 **Suggestion**

The AVX2 `_NOIBT_SUBROUTINE_SAFE` postcondition omits `read RSP s = word_add stackpointer (word 8)`, which is asserted both in the scalar `SHA3_KECCAK_F1600_NOIBT_SUBROUTINE_SAFE` (sha3_keccak_f1600.ml:894) and in every correctness theorem including this function's own `SHA3_KECCAK4_F1600_NOIBT_SUBROUTINE_CORRECT` (line 1097). Since `RSP` is in the MAYCHANGE frame this is sound but strictly weaker than the correctness proof — a caller cannot conclude the stack was properly unwound. Consider adding the `read RSP s = word_add stackpointer (word 8)` conjunct to match the scalar safety theorem and the correctness theorems for consistency, unless the manual lift makes proving it prohibitively costly (in which case a comment explaining the omission would help).

---

x86/proofs/sha3_keccak4_f1600_alt.ml
```diff
@@ -1309,3 +1309,403 @@ let SHA3_KECCAK4_F1600_ALT_WINDOWS_SUBROUTINE_CORRECT = prove
...
+           ==> ensures x86
+               (\s.
+                    bytes_loaded s (word pc)
+                    (BUTLAST sha3_keccak4_f1600_alt_tmc) /\
+                    read RIP s = word (pc + 14) /\
+                    read RSP s = stackpointer /\
+                    C_ARGUMENTS [bitstate_in; rc_pointer; rho8_ptr; rho56_ptr] s /\
+                    read events s = e)
+               (\s.
+                    read RIP s = word (pc + 2517) /\
```
aqjune-aws, line 1349: 🟡 **Suggestion**

Same observation as for `sha3_keccak4_f1600`: none of the four AVX2 `_alt` subroutine `_SAFE` theorems assert `read RSP s = word_add stackpointer (word 8)` in their postcondition, whereas the corresponding correctness theorems (e.g. SHA3_KECCAK4_F1600_ALT_NOIBT_SUBROUTINE_CORRECT at line 1030) and the scalar safety proof do. Consider restoring the RSP conjunct for parity so the safety spec is not weaker than the correctness spec on stack unwinding.

---

x86/proofs/sha3_keccak4_f1600.ml
```diff
@@ -1353,3 +1353,381 @@ let SHA3_KECCAK4_F1600_WINDOWS_SUBROUTINE_CORRECT = prove
...
+(* X86_PROMOTE_RETURN_STACK_TAC / WINDOWS_X86_WRAP_STACK_TAC do not apply.   *)
+(* Instead, the subroutine-level safety theorems are lifted manually from    *)
+(* the core SAFE theorem, mirroring the delta / word_and realignment dance   *)
+(* used by the correctness proofs above.                                     *)
+(* ------------------------------------------------------------------------- *)
+
+needs "x86/proofs/consttime.ml";;
+needs "x86/proofs/subroutine_signatures.ml";;
+
+let full_spec,public_vars = mk_safety_spec
```
aqjune-aws, line 1371: 🟢 **Nit**

`full_spec` is bound by `mk_safety_spec` here but never used (the theorem statement is written out by hand and proved directly via `PROVE_SAFETY_SPEC_TAC`). The scalar file uses `ASSERT_CONCL_TAC full_spec` (sha3_keccak_f1600.ml:872) to assert the hand-written statement matches the generated one, which is a valuable cross-check that guards against the hand-written spec drifting from what the signature/correctness pair implies. Consider either adding `ASSERT_CONCL_TAC full_spec` before `PROVE_SAFETY_SPEC_TAC` here (and in sha3_keccak4_f1600_alt.ml), or binding to `_,public_vars` to make the unused binding explicit.

---

x86/proofs/sha3_keccak4_f1600_alt.ml
```diff
@@ -1309,3 +1309,403 @@ let SHA3_KECCAK4_F1600_ALT_WINDOWS_SUBROUTINE_CORRECT = prove
...
+(* from the core SAFE theorem (the delta / word_and dance, as in the         *)
+(* correctness proofs above). It also reads the read-only rho8/rho56 rotate  *)
+(* tables; these are listed as input buffers in subroutine_signatures, so    *)
+(* mk_safety_spec treats them as public arguments, exactly like rc.          *)
+(* ------------------------------------------------------------------------- *)
+
+needs "x86/proofs/consttime.ml";;
+needs "x86/proofs/subroutine_signatures.ml";;
+
+let full_spec,public_vars = mk_safety_spec
```
aqjune-aws, line 1328: 🟢 **Nit**

As in sha3_keccak4_f1600.ml, `full_spec` is bound but unused here; the AVX2 `_alt` SAFE proof does not use `ASSERT_CONCL_TAC full_spec` to validate that the hand-written `SHA3_KECCAK4_F1600_ALT_SAFE` statement matches the auto-generated spec (which is where the newly-added rho8/rho56 public-arg handling actually gets exercised end-to-end). Adding the assertion here would give the most direct verification that the collect-signatures.py buffer change produces the intended public-args spec.

---

tools/collect-signatures.py
```diff
@@ -433,18 +433,24 @@ def checkOnlyInArch(fnname, onlyIn):
...
 
-    # input and output buffers
+    # input and output buffers.
+    # Buffer info comes from the single (shared) s2n-bignum.h comment, but a few
+    # functions take different arguments per arch (e.g. sha3_keccak4_f1600_alt
+    # has rho8/rho56 on x86 only). Skip buffers whose argument is absent from
+    # this arch's signature.
+    def emit_buffers(buffers):
+      for argname, bufferlen in buffers:
+        if argname not in arg_elem_bytesizes:
```
aqjune-aws, line 443: 🟡 **Suggestion**

The skip condition tests `argname not in arg_elem_bytesizes`, but that dict is populated only for pointer/array-typed args (the loop at lines 420-434 `continue`s for scalar args). So a buffer entry whose argument exists in this arch's signature but is a non-pointer scalar type would be silently skipped rather than emitted. That cannot happen today (buffers only come from pointer/array args), so this is correct in practice, but keying the "absent from this arch" test off `arg_elem_bytesizes` conflates two distinct conditions ("arg absent" vs "arg present but not a buffer type"). Consider testing against the actual per-arch arg-name set (e.g. build `argnames = {a[0] for a in fnsig.args}` and check `argname not in argnames`) so the intent — skip buffers for absent args — is expressed directly and a future non-pointer buffer arg would surface as the intended KeyError rather than being dropped.

---

x86/proofs/sha3_keccak4_f1600.ml
aqjune-aws: My evaluation of AI comment: it is fine to omit `read RSP s = word_add stackpointer (word 8)` in the `_SAFE` theorems because they already exist in the `_CORRECT`s. Probably https://github.com/awslabs/s2n-bignum/blob/main/doc/s2n_bignum_pr_checklist.md should be updated to mention this.

x86/proofs/sha3_keccak4_f1600_alt.ml
aqjune-aws: My evaluation of AI comment: same as the previous one

x86/proofs/sha3_keccak4_f1600.ml
aqjune-aws: My evaluation of AI comment: Yeah, this seems like a valid comment.

x86/proofs/sha3_keccak4_f1600_alt.ml
aqjune-aws: My evaluation of AI comment: same as the previous one

tools/collect-signatures.py
aqjune-aws: My evaluation of AI comment: this seems to make sense. Maybe add the sanity check here?

PR #421, opened 2026-07-02
x86: Add VPBROADCASTW instruction model

jakemas:
Adds a model of the x86-64 AVX/AVX2 `VPBROADCASTW` instruction (VEX.128/256.66.0F38.W0 0x79),
following the convention in `x86/INSTRUCTION.md` (the same 5 areas as e.g. VPMULHUW #… /
VPBROADCASTD, which this mirrors).

Needed to model an ML-DSA AVX2 rejection-sampling routine that broadcasts a 16-bit
constant.

## What changed (5 files)

- **`x86/proofs/decode.ml`** — decode the VEX `66 0F38 79 /r` form to `VPBROADCASTW`
  (reg or 16-bit memory source; 128- or 256-bit destination via the VEX.L bit; `W0`).
  Distinct opcode branch (`0x79`) from the existing `VPBROADCASTD` (`0x58`) — disjoint.
- **`x86/proofs/instruction.ml`** — `VPBROADCASTW operand operand` constructor.
- **`x86/proofs/x86.ml`** — `x86_VPBROADCASTW` semantics (broadcast the low 16-bit word of
  the source across the destination via `word_duplicate`), the `x86_execute` dispatch arm
  (load/store events + 128/256-bit dest × 16/128-bit src operand sizes), the
  `x86_VPBROADCASTW_ALT` `EXPAND_SIMD_RULE`, and its registration in `X86_OPERATION_CLAUSES`.
- **`x86/proofs/simulator.ml`** — 8 register-to-register cosimulation entries (xmm/ymm
  destination variants, mirroring the `VPBROADCASTD` block).
- **`x86/allowed_asm`** — `: vpbroadcastw$` (constant-time; alphabetical, after `vpbroadcastq`).

## Validation

- Decode verified: the real byte forms `c4 e2 7d 79 f6` / `c4 e2 79 79 c1` / etc. decode to
  `VPBROADCASTW (ymm6)(xmm6)` / `(xmm0)(xmm1)` etc. via `X86_MK_EXEC_RULE`.
- `x86/proofs/base.ml` loads cleanly with the change (no regression on the full model re-typecheck).
- Semantics model mirrors the existing, tested `VPBROADCASTD` broadcasting a smaller element.
- Cosimulation (`make sematest`) exercises the new simulator entries against hardware — runs in CI.


jargh (approved):
All perfect, thank you - I will merge.

PR #424, opened 2026-07-10
Fix some static analysis findings

jargh:
*Issue #, if available:*

*Description of changes:*


jakemas (commented):
Went through the whole diff. Mostly spelling/comment fixups which all look fine. Couple of real changes worth calling out, and I checked each one:

mldsa_reduce: nice — the old `MLDSA_REDUCE_CORRECT` only said `output = mldsa_barred(input)`, now it actually proves congruence mod `8380417` + the centered bound, under the `a <= 0x7fbfffff` precondition (otherwise the internal `+4194304` overflows). Precondition is carried through both subroutine wrappers and documented in the `.S`/header. Reads clean.

keccak4: dropping the dead `CMP X25, #23` is a good catch, and the proof side tracks it properly — offsets (`0xc38`->`0xc30`, exit `0xc08`->`0xc00`), branch displacements and step counts all shifted by the 4 bytes. Same across all three variants.

aes_xts_decrypt (`subs`->`sub`): fine in both spots — the first is followed by `cmp x2, #16` and the second by `subs x8, #1`, so the flags are dead before anything reads them. Bytecode in the `.ml` matches.

sm2 comment fix (`p_256` -> `p_sm2`) is correct too.

Looks good to me!

jakemas (commented):
Also reviewed [193c638](https://github.com/awslabs/s2n-bignum/pull/424/commits/193c6382fcaedb0d6920c547999291bafd41553d) after my first review. All good

jakemas (approved):

PR #423, opened 2026-07-08
Add AArch64 HOL-Light proofs for six ML-DSA polynomial routines

jakemas:
#### summary

Ports the AArch64 HOL-Light proofs for six ML-DSA polynomial routines from the [mldsa-native](https://github.com/pq-code-package/mldsa-native) project into s2n-bignum, bringing their formal functional-correctness and memory-safety proofs under the s2n-bignum proof framework. The routines:

- `mldsa_caddq`
- `mldsa_chknorm`
- `mldsa_decompose_32/88`
- `mldsa_polyz_unpack_17/19`

simulator_iclasses.ml: adds the no-offset LD1/ST1 (2-register, 128-bit) encodings to skipping_iclasses so check_insns() (make sematest) recognizes them — they appear in the new object code for the first time but reuse existing decode/semantics, so no decode.ml/instruction.ml change is needed.

PR #426, opened 2026-07-11
Add x86 HOL-Light proofs for seven ML-DSA polynomial routines

jakemas:
Adds x86-64 (AVX2) HOL Light proofs for five ML-DSA polynomial routines, ported from mldsa-native, mirroring the existing merged `x86/proofs/mldsa_caddq.ml` convention:

- `mldsa_chknorm` — infinity-norm bound check
- `mldsa_decompose_32` — coefficient decomposition, GAMMA2 = (Q-1)/32 (ML-DSA-65/87)
- `mldsa_decompose_88` — coefficient decomposition, GAMMA2 = (Q-1)/88 (ML-DSA-44)
- `mldsa_polyz_unpack_17_x86` — unpack packed z, GAMMA1 = 2^17 (ML-DSA-44)
- `mldsa_polyz_unpack_19_x86` — unpack packed z, GAMMA1 = 2^19 (ML-DSA-65/87)

Each routine gets the full x86 variant set: SysV `CORRECT` / `NOIBT_SUBROUTINE_CORRECT` / `SUBROUTINE_CORRECT`, the constant-time/memory-safety `SAFE` variants, and the Windows-ABI `_WINDOWS_SUBROUTINE_CORRECT` / `_SAFE` variants, all proven with no added axioms.

The `polyz_unpack` routines are suffixed `_x86` because the x86 implementation uses a different ABI from the aarch64 one (it unpacks directly from the byte buffer with inline constants, and so takes no shuffle-table argument).

Includes the Intel-syntax `x86/mldsa/*.S`, the AT&T `x86_att/mldsa/*.S` (with a small extension to `x86_att/attrofy.sed` to cover the new macros and the `al`/`bl`/`dl` byte registers), Makefile object entries, `subroutine_signatures.ml` / `specifications.txt` entries, and benchmark/test wiring.

Based on the aarch64 PR; the diff here is the x86 addition on top of it.

Note: this is stacked on top of #423 (the aarch64 proofs); until that merges, the diff here includes the #423 commit as well. The x86 work is the top commit (`Add x86 HOL-Light proofs for five ML-DSA polynomial routines`).


jargh (approved):
All great, thank you! I'll merge and all the MLDSA work in s2n-bignum will be complete!

PR #427, opened 2026-07-13
Relax unnecessary ML-KEM alignment preconditions

BenKnill:
## Summary

This removes two alignment assumptions that are stronger than the x86 ML-KEM
implementations require. `mlkem_frombytes` uses `vmovdqu` for loads from input
`a`, so only its `vmovdqa` output stores require 32-byte alignment.
`mlkem_tobytes` has the opposite pattern: its input loads use `vmovdqa`, while
its output stores use `vmovdqu`.

The redundant assumption is removed consistently from the core, no-IBT, SysV
subroutine, Windows no-IBT, and Windows subroutine theorem layers. There is no
assembly or C API/ABI change; the formal theorem contracts are strengthened.

## Validation

- Exact full-source warm replays observed all extracted theorem probes for both
  edited proof files.
- A static proof/assembly contract audit reports no remaining formal-only
  alignment parameter for these functions.
- The warm replay does not bind a fresh dependency basis, so this is a draft
  pending the repository's full x86 proof CI. No security vulnerability is
  claimed.

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.


jargh:
Good catch, thanks! If you "undraft" the PR we can probably merge it pretty soon since it's a clear, localized improvement.

jargh:
Happy to merge this soon - if you want to switch the PR from "draft" to "ready for review" and rebase it, that would be great, otherwise I'll just do it all.

jargh (approved):
Thanks again!

PR #430, opened 2026-07-24
Arm: Decode by-element SMLAL, UMULL, and UMLSL

ckocaogullar:
Adds indexed-lane decoding and hardware cosimulation coverage for `SMLAL`, `UMULL`, `UMLSL`, and their `2` forms. These instructions multiply each 16- or 32-bit `Vn` lane by a selected `Vm` lane, and produce 32- or 64-bit results:

  - `SMLAL` performs a signed multiply and adds the product to the wide destination.
  - `UMULL` performs an unsigned multiply and writes the wide product.
  - `UMLSL` performs an unsigned multiply and subtracts the product from the wide destination.

The selected `Vm` lane is reused across all SIMD lanes. Base forms use the low half of `Vn`; `2` forms use the high half. These instructions are needed by the x265 AArch64 NEON proof work.

 ### Implementation
The decoder reuses the existing widening vector semantics by reading the indexed `Vm` lane through `QLANE`, which duplicates it across the SIMD lanes. `Q` selects the base or `2` form, and the undefined `sz = 00` and `sz = 11` encodings decode to `NONE`.

  ### Validation
  - Checked all 12 valid family, element-size, and `Q` combinations against LLVM, and decoded each representative in HOL.
  - Confirmed that six undefined-size representatives decode to `NONE` and that the new patterns do not overlap existing instruction classes.
  - Full 16-worker AArch64 `make -C arm sematest` passed 53,014 cases with zero failures, including 356 valid indexed-lane executions and 350 undefined-size rejections.


jargh (approved):
All looks great - I've just rebased it

PR #431, opened 2026-07-24
Arm: Model SADDW, SSUBL, SSUBW, USUBL, and USUBW

ckocaogullar:
Adds semantics, decoding, conversion registration, and hardware cosimulation coverage for `SADDW`, `SSUBL`, `SSUBW`, `USUBL`, `USUBW`, and their `2` forms. These instructions perform lane-wise arithmetic using 8-, 16-, or 32-bit narrow elements and 16-, 32-, or 64-bit destination elements:

  - `SADDW` adds sign-extended narrow `Vm` lanes to already-wide `Vn` lanes.
  - `SSUBL` and `USUBL` widen both narrow operands before computing `Vn - Vm`.
  - `SSUBW` and `USUBW` widen narrow `Vm` lanes and subtract them from the already-wide `Vn` lanes.

The signed forms use sign extension; the unsigned forms use zero extension.

These instructions are needed by the x265 AArch64 NEON proof work.

  ### Implementation
`Q = 0` selects the base form and reads each narrow operand from the low half of its source register. `Q = 1` selects the corresponding `2` form and reads the high half. The decoder supports 8-, 16-, and 32-bit narrow elements and rejects the undefined `size = 11` encoding.

The new semantics are registered as expanded SIMD rules. The simulator masks cover every family, source size, and base/`2` form, including the undefined size encodings.

  ### Validation
  - Assembled all 30 valid family, source-size, and `Q` combinations with LLVM and checked their opcode fields.
  - Fresh HOL checks decoded all 30 valid representatives and confirmed that ten `size = 11` representatives decode to `NONE`.
  - Confirmed that the decoder clauses and simulator masks do not overlap existing instruction classes.
  - Full 16-worker AArch64 `make -C arm sematest` passed 54,079 cases with zero failures, including 1,857 valid executions and 586 undefined-size rejections.

### DIFF
--- a/arm/proofs/decode.ml
+++ b/arm/proofs/decode.ml
@@ -1018,6 +1018,32 @@ let decode = new_definition `!w:int32. decode w =
       else
         SOME (arm_UMLSL_VEC (QREG' Rd) (QREG' Rn) (QREG' Rm) (val esize))
 
+  | [0:1; q; 0b101110:6; size:2; 0b1:1; Rm:5; 0b001000:6; Rn:5; Rd:5] ->
+    // USUBL (vector, Q = 0). USUBL2 (vector, Q = 1)
+    // Unsigned widening long subtract: both source operands are narrow
+    // (esize-bit) lanes that get zero-extended to 2*esize.
+    // size=11 (esize=64) is UNDEFINED.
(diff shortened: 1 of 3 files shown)


jargh (approved):
All great, thanks - I will merge.

PR #432, opened 2026-07-28
Arm: Model narrowing shifts: SQSHRUN, SQRSHRUN, and RSHRN

ckocaogullar:
Adds semantics, decoding, conversion registration, and hardware cosimulation coverage for `SQSHRUN`, `SQRSHRUN`, `RSHRN`, and their `2` forms. These instructions shift 16-, 32-, or 64-bit vector lanes right and narrow them to 8-, 16-, or 32-bit results:

  - `SQSHRUN` performs a signed arithmetic right shift and saturates the result into the unsigned destination range.
  - `SQRSHRUN` performs the same signed-to-unsigned narrowing with rounding before the shift.
  - `RSHRN` rounds and shifts right, then truncates the result without saturation.

Base forms write the narrowed result to the low destination half and clear the high half. The `2` forms preserve the low half and write the result to the high half. These instructions are needed by the x265 AArch64 NEON proof work.

  ### Implementation

The decoder derives the destination element size and shift amount from `immh:immb`, while `Q` selects the base or `2` form. Reserved 128-bit source encodings decode to `NONE`.

`SQSHRUN` is decoded inside the existing Advanced SIMD modified-immediate clause so it does not shadow valid immediate encodings. The `SQRSHRUN` and `RSHRN` encodings are disjoint from that region. The new semantics are registered as expanded SIMD rules, and the simulator masks cover valid sizes and shifts as well as reserved encodings.

  ### Validation

  - Assembled all 336 valid family, form, element-size, and shift combinations with LLVM and checked their opcode fields.
  - Fresh HOL checks decoded all of those 336 valid encodings, confirmed that 12 reserved encodings decode to `NONE`, preserved two BIC-immediate regressions, and checked eight saturation and rounding boundaries.
  - Confirmed that the new decoder clauses and simulator masks do not overlap existing instruction classes.
  - Full 16-worker AArch64 `make -C arm sematest` passed 52,487 cases with zero failures, including 890 valid executions and 502 reserved-encoding rejections.


jargh (approved):
All looks great! Some of the helper functions that round (`word_uqrshrun` and `word_rshrn`) degenerate wrongly, or at least arguably so, in the case shift=0 (because of cutoff subtraction in shift-1), which is something to note if these ever get used elsewhere. But by construction this never happens from the decoding in these instructions.

PR #434, opened 2026-07-28
Arm: Model XTN2, three-register TBL, 64-bit EXT, and DUP (element)

ckocaogullar:
Adds semantics, decoding, conversion registration, and hardware cosimulation coverage for `XTN2`, three-register `TBL`, 64-bit `EXT`, and `DUP` (element).

`XTN2` narrows each source lane and writes the results into the high half of the destination while preserving its low half. Three-register `TBL` looks up bytes in a 48-byte table formed from three consecutive vector registers. The 64-bit `EXT` form extracts an eight-byte window from two concatenated source registers, and `DUP` (element) broadcasts one source lane across the destination.

These forms are needed by x265 AArch64 NEON proof work.

  ### Implementation

`XTN2` extends the existing `XTN` decoder, with `Q` selecting the base or high-half form. Three-register `TBL` models register-list wraparound modulo 32 and returns zero for indices outside its 48-byte table. The 64-bit `EXT` form reuses `arm_EXT` through `DREG'` views and rejects `imm4[3] = 1`.

`DUP` (element) derives the element size from the least significant set bit of `imm5` and the source lane index from the remaining bits. It supports legal 64- and 128-bit destinations and rejects reserved immediate combinations.

  ### Validation

  - Checked all 71 new legal forms and 19 existing `XTN`/`EXT` forms against LLVM, and confirmed that LLVM rejects 15 reserved encodings.
  - Fresh HOL checks decoded all 71 new forms, preserved the 19 existing forms, proved the 15 reserved encodings decode to `NONE`, and checked 12 semantic boundaries with `axioms() = 3`.
  - Confirmed that the changed decoder clauses and five simulator masks do not overlap existing decoder clauses or simulator classes.
  - Full 64-worker AArch64 `make -C arm sematest` passed 366,749 cases with zero failures, including 5,147 valid executions covering all 71 new forms and 1,385 reserved-encoding rejections. Three-register `TBL` exercised all 32 table-start registers, including 116 wraparound
  executions.

### DIFF
--- a/arm/proofs/decode.ml
+++ b/arm/proofs/decode.ml
@@ -1092,12 +1092,13 @@ let decode = new_definition `!w:int32. decode w =
       let esize: (64)word = word_shl (word 8: (64)word) (val size) in
       SOME (arm_UZP2 (QREG' Rd) (QREG' Rn) (QREG' Rm) (val esize))
 
-  | [0:1; 0:1; 0b001110:6; size:2; 0b100001001010:12; Rn:5; Rd:5] ->
-    // XTN
+  | [0:1; q; 0b001110:6; size:2; 0b100001001010:12; Rn:5; Rd:5] ->
+    // XTN (q=0, low half) / XTN2 (q=1, high half)
     if size = (word 0b11: (2)word) then NONE // "UNDEFINED"
     else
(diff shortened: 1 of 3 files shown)


jargh (approved):
All great, thank you!

PR #435, opened 2026-07-28
Arm: Add TBZ and TBNZ

ckocaogullar:
Adds AArch64 model support for `TBZ` and `TBNZ`, including decoding, semantics, symbolic-execution rules, and simulator bookkeeping.

`TBZ` branches when a selected register bit is zero, while `TBNZ` branches when it is nonzero. These instructions are needed by the x265 AArch64 proof work.

### Implementation

The decoder reconstructs the bit index from `b5:b40`. `b5 = 0` selects a 32-bit `W` register and bits 0 through 31; `b5 = 1` selects a 64-bit `X` register and bits 32 through 63. The signed `imm14` field supplies a four-byte-scaled PC-relative branch offset.

The semantics model both taken and fall-through paths using the existing `PC` and `EventJump` conventions, and `_ALT` rules register both instructions for symbolic execution. Both encoding classes are added to `skipping_iclasses`, matching the existing treatment of control-flow instructions.

  ### Validation

  - Checked 896 encodings against LLVM across both instructions, all 64 bit indices, and seven offsets including both signed boundaries; LLVM rejected five invalid assembly forms.
  - Fresh HOL checks decoded the same 896 encodings and both target x265 encodings, preserved 20 `CBZ`/`CBNZ` decodes, proved eight taken/not-taken semantic cases, exercised both operation clauses, and reported `axioms() = 3`.
  - Confirmed that the decoder and simulator masks do not overlap existing classes and uniquely route 2,560 representative encodings.
  - The native simulator built against HOL Light `b19af6934759eed622d7aca25092367f81656eee`. All 64 AArch64 sematest workers completed, covering 326,182 register and 36,212 memory instances (362,394 total) with zero failures. Because `TBZ` and `TBNZ` are in `skipping_iclasses`,
  sematest is a repository-wide regression check and does not execute them; the LLVM and HOL checks provide direct coverage.


jargh (approved):
All great! This one won't be exercised by the simulator harnesses at the moment, but the close analogy with CBZ/CBNZ makes the situation quite reassuring.

PR #436, opened 2026-07-30
Arm: Add four-register LD1 and ST1

ckocaogullar:
Adds semantics, decoding, conversion registration, and hardware cosimulation coverage for the four-register `LD1` and `ST1` multiple-structure forms.

`LD1` loads four consecutive SIMD registers from contiguous memory, while `ST1` stores them. The 64-bit `D` forms transfer 32 bytes and the 128-bit `Q` forms transfer 64 bytes. These forms are needed by the x265 AArch64 proof work.

### Implementation

The decoder supports all four element-size fields with no offset, post-immediate writeback (`#32` for `D` and `#64` for `Q`), and post-register writeback. Register lists wrap modulo 32, so a list beginning at `V29` continues through `V0`.

`arm_LDP4` and `arm_STP4` model the operation as four contiguous little-endian loads or stores and record the complete 32- or 64-byte `EventLoad` or `EventStore` span. A bounded memory generator exercises every addressing mode safely against hardware.

### Validation

  - Checked all 48 load/store, register-width, addressing-mode, and size-field combinations against LLVM using wrapped register lists and SP bases; LLVM rejected two incorrect post-immediate amounts.
  - Fresh HOL checks decoded the same 48 representatives, preserved ten nearby load/store decodes, exercised both load/store clauses, confirmed the 32- and 64-byte event spans, and reported `axioms() = 3`.
  - Confirmed that the four decoder clauses and eight simulator masks do not overlap existing instruction classes and that every legal combination maps to exactly one decoder and simulator class.
  - Full four-worker native AArch64 `make -C arm sematest` passed 12,067 register and 1,343 memory cases (13,410 total) with zero failures. It directly executed 149 new forms and sampled 44 of the 48 field combinations; the exhaustive LLVM and HOL checks cover all 48.

### DIFF
--- a/arm/proofs/decode.ml
+++ b/arm/proofs/decode.ml
@@ -91,6 +91,18 @@ let arm_ldstp_q = new_definition `arm_ldstp_q ld Rt Rt2 =
 let arm_ldstp_2q = new_definition `arm_ldstp_2q ld Rt =
   let Rtt:(5 word) = word ((val Rt + 1) MOD 32) in
   (if ld then arm_LDP else arm_STP) (QREG' Rt) (QREG' Rtt)`;;
+let arm_ldstp_4q = new_definition `arm_ldstp_4q ld Rt =
+  let Rt2:(5 word) = word ((val Rt + 1) MOD 32) in
+  let Rt3:(5 word) = word ((val Rt + 2) MOD 32) in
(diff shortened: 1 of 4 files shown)


jargh (approved):
All looks very good - thanks for also generalizing the cosimulation harness, which makes me feel much more confident about load/store instructions. I will merge.

PR #437, opened 2026-07-30
Arm: Model SIMD LD1/ST1, LD2/ST2, LDR/STR, and LDUR/STUR forms

ckocaogullar:
Adds instruction models, decoding, conversion support, and hardware cosimulation coverage for these AArch64 SIMD load/store forms:

- Three-register `LD1` and `ST1` for consecutive `D` or `Q` registers, with no offset, post-immediate writeback, or post-register writeback.
- No-offset, two-register `LD2` and `ST2`, which deinterleave loads and interleave stores.
- Single-lane `LD1` and `ST1` for 32-bit SIMD lanes, with no offset, `#4` post-immediate writeback, or post-register writeback.
- Scalar SIMD `LDR` and `STR` for `B`, `H`, and `S` registers with scaled unsigned offsets.
- Scalar SIMD `LDUR` and `STUR` for `S` and `D` registers with signed unscaled offsets.

These forms are needed by the x265 AArch64 proof work.

### Implementation

The three-register forms transfer contiguous 24- or 48-byte register lists, including lists that wrap from `V31` to `V0`. For post-indexed forms, `Rm = 31` advances the base by `#24` or `#48`; other `Rm` values select register writeback through `Xm`.

The `LD2`/`ST2`, single-lane, and scalar forms reuse the existing SIMD load/store semantics. New `B`, `H`, and `S` component expansions expose the complete `Q`-register state to symbolic execution.

### Validation

- LLVM 11 accepted 114 legal representatives and rejected ten invalid forms.
- Fresh HOL checks decoded all 114 legal representatives, proved five reserved forms decode to `NONE`, and confirmed that `ARM_MK_EXEC_RULE` loads the reported `0x4cdf6c34` instruction.
- Confirmed that the 13 new decoder clauses and seven simulator masks do not overlap existing instruction classes.
- Full 64-worker native AArch64 `make -C arm sematest` passed 368,780 cases with zero failures. This included 2,817 three-register `LD1`/`ST1` executions covering all 48 width, direction, arrangement, and addressing-mode combinations.

### DIFF
--- a/arm/proofs/arm.ml
+++ b/arm/proofs/arm.ml
@@ -251,6 +251,36 @@ let DREG_EXPAND_CLAUSES = prove
    D31 = Q31 :> zerotop_64`,
   REWRITE_TAC(!component_alias_thms));;
 
+(* Expand the scalar SIMD&FP sub-registers (S/H/B) down to the full Q register
+   (using the Qn aliases, so the result matches the Q-register state exposed by
+   the cosimulator's regfile) composed with the appropriate zerotop chain.  This
+   lets reads/writes of the S/H/B components reduce against the Q-register state
+   in the same way the existing WREG/DREG expand clauses do.  The decoder
+   produces SREG n / HREG n / BREG n with a concrete index n, so we enumerate
+   the 32 registers; the Qn alias on the RHS is what the input/regfile rewrite
+   knows about. *)
+
+let SREG_EXPAND_CLAUSES = prove
+ (list_mk_conj (map (fun i ->
+    let s = string_of_int i in
+    parse_term("SREG "^s^" = (Q"^s^" :> zerotop_64) :> zerotop_32")) (0--31)),
+  REWRITE_TAC[SREG; DREG] THEN REWRITE_TAC(!component_alias_thms));;
+
+let HREG_EXPAND_CLAUSES = prove
+ (list_mk_conj (map (fun i ->
+    let s = string_of_int i in
+    parse_term("HREG "^s^" = ((Q"^s^" :> zerotop_64) :> zerotop_32) :> zerotop_16"))
+    (0--31)),
+  REWRITE_TAC[HREG; SREG; DREG] THEN REWRITE_TAC(!component_alias_thms));;
+
+let BREG_EXPAND_CLAUSES = prove
+ (list_mk_conj (map (fun i ->
+    let s = string_of_int i in
+    parse_term("BREG "^s^
+      " = (((Q"^s^" :> zerotop_64) :> zerotop_32) :> zerotop_16) :> zerotop_8"))
+    (0--31)),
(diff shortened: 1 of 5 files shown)


nebeid:
Possible gap, spotted while proving aws-lc's `gcm_ghash_v8`: this PR looks to add
only the **no-offset** 3-register `LD1`/`ST1` arms — `0b0011000` with opcode
`0b0110`, for both Q and D — and I couldn't find a **post-index** 3-register arm
(`0b0011001`, i.e. `Postimmediate_Offset (word 48)` for the Q form) anywhere in
the diff. #436 does include the post-index 4-register form, so this looks
inadvertent rather than deliberate.

It shows up in practice: `gcm_ghash_v8`'s 4x prologue does

```
ld1 {v20.2d,v21.2d,v22.2d},[x1],#48     // 0x4cdf6c34
```

to load `H`, the packed Karatsuba mids and `H^2` in one go, so the function can't
be loaded without it — `ARM_MK_EXEC_RULE` raises on the unmodelled word rather
than degrading.

For what it's worth, I had independently written the same models before finding
this series, and the overlapping parts are identical to yours — `arm_LDP3`,
`arm_STP3`, the `arm_ldstp_3q` helper and the compset registration all match
character for character, which is reassuring for both of us. Mine adds the arm as

```
  | [0:1; 1:1; 0b0011001:7; is_ld; 0:1; 0b11111:5; 0b0110:4; size:2; Rn:5; Rt:5] ->
    SOME (arm_ldstp_3q is_ld Rt (XREG_SP Rn) (Postimmediate_Offset (word 48)))
```

though in your style it would presumably bind `Rm` and branch on `val Rm = 31` to
pick up post-register too, as #436 does for the 4-register case.

I also ran a co-simulation pass restricted to the 3-/4-register forms on a
Graviton host — 86 instances, both register counts, both addressing modes,
unaligned bases and wrapping register lists (e.g. a list starting at `V29`), all
matching the model. Happy to hand that generator over if it's useful.

ckocaogullar:
Thanks for the heads up @nebeid ! Seems like I only focused on what I immediately needed and overlooked the modes you mention. I'm working on adding them now, I'll update once it's ready.

ckocaogullar:
Finally had time to get back to this, I’ve added the post-indexed three-register `LD1`/`ST1` forms for both Q and D registers. `Rm = 31` now selects `#48` or `#24` writeback, while other `Rm` values select post-register writeback through `Xm`.

I also extended the cosimulation generator to cover no-offset, post-immediate, and post-register forms. `ARM_MK_EXEC_RULE` now accepts `0x4cdf6c34`, and the full CI and 64-worker AArch64 sematest pass.

Let me know if there's anything that needs changing.

nebeid (approved):

nebeid (approved):

PR #442, opened 2026-08-17
Arm: Decode two-register LD1 and ST1 post-register forms

ckocaogullar:
Adds post-register writeback decoding and hardware cosimulation coverage for the 128-bit, two-register `LD1` and `ST1` multiple-structure forms.

These instructions transfer two consecutive SIMD registers between memory and the register file, for 32 bytes total. The post-register form uses `Xn` as the memory address and then advances it by `Xm`. This form is needed by x265 AArch64 NEON proof work.

  ### Implementation

The decoder reuses the existing `arm_ldstp_2q` semantics. `Rm = 31` retains the existing post-immediate `#32` behavior; any other `Rm` selects `Postreg_Offset (XREG' Rm)`. Consecutive register lists continue to wrap modulo 32.

The simulator now exercises no-offset, post-immediate, and post-register forms for loads and stores across all four element arrangements. The instruction masks are widened to include register writeback.

  ### Validation

  - LLVM 11 assembled post-register and post-immediate loads and stores for all four element arrangements, producing the same 16 exact encodings used by the HOL checks.
  - Fresh HOL checks decoded all 16 representatives. `check_insns()` confirmed that the changed decoder clause and simulator masks do not overlap existing classes.
  - Focused AArch64 hardware cosimulation passed 144 cases, covering every load/store, element arrangement, and addressing-mode combination.
  - Full 64-worker AArch64 `make -C arm sematest` passed 328,908 register and 36,670 memory cases (365,578 total) with zero failures, including 1,509 post-register executions (739 loads and 770 stores).

### DIFF
--- a/arm/proofs/decode.ml
+++ b/arm/proofs/decode.ml
@@ -411,10 +411,12 @@ let decode = new_definition `!w:int32. decode w =
     SOME (arm_ldst_q is_ld Rt (XREG_SP Rn) No_Offset)
 
   // LD1/ST1 (multiple structures), 2 registers,
-  //   Post-index with immediate offset, datasize = 128
+  //   Post-immediate offset and post-register offset, datasize = 128
   // Similar to LDP of SIMD registers, assuming little-endian architecture.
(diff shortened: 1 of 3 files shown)


jargh (approved):
All good, thanks! I'll merge.

PR #445, opened 2026-08-19
ARM: whole-blocks-only AES-256-GCM decrypt kernel + HOL Light proof

nebeid:
### ARM: whole-blocks-only AES-256-GCM decrypt kernel + HOL Light proof

Adds `aesv8_gcm_8x_dec_256_wb`, an AES-256-GCM decrypt kernel for whole 16-byte blocks, together with its HOL Light proof, the shared AES-GCM specification/support layer it needs, test coverage, and benchmark wiring.

#### What changed

This kernel is derived from aws-lc's 8x-unrolled AES-256-GCM decrypt path with two contract-level changes:

1. It accepts only nonzero multiples of 128 bits. The entry guard `tst x1, #127; b.ne ...` returns 0 without touching memory otherwise.
2. It deletes the partial-final-block masking path. Under the whole-blocks contract that code is dead, so the kernel no longer does the original final-block overread or output read-modify-write.

This branch also includes:

- a fused 1-4 block fast path for the small-size whole-block cases
- code-size / footprint note for that fused path: the checked-in `aesv8_gcm_8x_dec_256_wb.o` `__text` size grows from `4976` to `5960` bytes (`+984`, `1.20x`) relative to the unfused whole-block kernel, while the stack frame remains `80` bytes and the register / proof-side `MAYCHANGE` footprint stays unchanged
- `.balign 16` before the main loop to pin the loop entry at `addr mod 16 = 0`
- proof-side use of `LENGTH aesv8_gcm_8x_dec_256_wb_mc` instead of a hardcoded code-size literal
- the small AES-XTS shared-spec renames needed to avoid HOL Light global-name collisions; those renames account for the `-117` lines in the diff

#### Proof

The PR exports:

- `AESV8_GCM_8X_DEC_256_CORRECT`
- `AESV8_GCM_8X_DEC_256_SUBROUTINE_CORRECT`

Both are proved with `hyps = 0`, `axioms = 3`, and the proof is stated in the shared AES-CTR/AES-GCM vocabulary so the encrypt and decrypt sides talk about the same counter/keystream model.

Within the `<=8` whole-block path:

- a single dispatch theorem covers `nblk = 1..8`
- cases `1..4` are now discharged by the fused small-path proofs `WB_FUSED_1BLOCK`, `WB_FUSED_2BLOCK`, `WB_FUSED_3BLOCK`, and `WB_FUSED_4BLOCK`
- cases `5..8` continue to use the existing band proofs
- that `<=8` dispatch theorem is then consumed by the main byte-list spine, which in turn feeds the exported `AESV8_GCM_8X_DEC_256_CORRECT` theorem
- so the fused `1..4` proofs are not standalone side lemmas; they are the actual `1..4` cases of the main `<=8` proof path

#### Testing

`tests/test.c` adds:

- a differential test against the pure-C aws-lc GCM reference
- a whole-blocks guard test that asserts return value 0 and no mutation of `out`, `Xi`, or `ivec`, including rejected lengths in the dispatch-relevant range
- known-answer tests using the aws-lc AES-256-GCM vector set
- filtered Wycheproof AES-256-GCM valid vectors that stay within this kernel's whole-block contract

#### Benchmarking

`benchmarks/benchmark.c` now registers this kernel at:

- 16, 32, 48, 64, 128, 256, 512, 1024, and 4096 bytes

Headline performance against current aws-lc `aesv8_gcm_8x_dec_256`:

- non-fused whole-block path: about `-19%` to `-22%` at 128 B, `-14%` to `-18%` at 256 B, and low-single-digit / near-floor deltas by 4 KB
- fused 1-4 block path: about `-21%` to `-51%` at 16-64 B depending on core, and effectively neutral above 64 B

Through the aws-lc wrapper, the measured 256 B gain is about:

- `-5.8%` on Graviton3
- `-6.1%` on Graviton4
- `-8.0%` on Graviton5

#### Follow-up outside this PR

This is still the s2n-bignum side only. Importing it into aws-lc will need the usual coordination on:

- symbol/caller wiring
- removal of the duplicate perlasm definition on the aws-lc side
- the dispatch-threshold decision for exposing the 128 B and fused-path wins


nebeid:
## Baseline and fused-path benchmarks

- Cross-core kernel comparisons against AWS-LC and the pre-optimization kernel.
- Fused 1-4-block results and the initial AWS-LC wrapper measurement.

---

Preserving the benchmark tables from the earlier draft here so the PR body can stay short.

All figures below come from one binary per host holding every variant plus an A/A twin for each, timed round-robin interleaved, `taskset`-pinned, 30 processes, estimator = median of per-process paired deltas with bootstrap CIs. Negative = this variant is faster. `~` marks a delta inside that size's A/A floor, i.e. not resolved.

**Versus current aws-lc `aesv8_gcm_8x_dec_256`.**

| size | V1 | V2 | V3 |
|---:|---:|---:|---:|
| 16 B | −7.20 % | −0.61 %~ | −10.18 % |
| 32 B | −8.94 % | −13.08 % | −9.91 % |
| 48 B | −7.57 % | −16.02 % | −7.96 % |
| 64 B | −8.25 % | −16.29 % | −7.31 % |
| 128 B | −22.40 % | −21.58 % | −19.13 % |
| 256 B | −16.30 % | −17.59 % | −14.21 % |
| 512 B | −9.45 % | −10.21 % | −9.69 % |
| 1024 B | −4.65 % | −4.95 % | −5.38 % |
| 4096 B | −0.37 % | −1.60 % | −1.36 % |

**How much of that the optimisation commits account for.**

This is the same kernel measured against the whole-blocks kernel before the five optimisation commits:

| size | V1 | V2 | V3 |
|---:|---:|---:|---:|
| 16 B | −4.49 % | +2.37 % | −6.21 % |
| 32 B | −3.38 % | −8.08 % | −5.82 % |
| 48 B | −3.61 % | −12.27 % | −4.54 % |
| 64 B | −5.67 % | −12.23 % | −4.83 % |
| 128 B | −22.19 % | −20.64 % | −17.33 % |
| 256 B | −16.25 % | −16.73 % | −13.94 % |
| 512 B | −9.55 % | −10.32 % | −9.58 % |
| 1024 B | −4.86 % | −5.63 % | −5.36 % |
| 4096 B | −0.76 % | −1.71 % | −1.37 % |

The gains concentrate at 128 B and amortise away by 4 KB. The region `.L256_dec_main_loop:` to `.L256_dec_tail:` is byte-identical across the six optimisation commits, so the measured movement comes from setup/tail work and code layout, not from changes inside the bulk loop.

**Fused 1-4 whole-block path.**

Measured against the same reference kernel, same binary and method:

| size | V1 | V2 | V3 |
|---:|---:|---:|---:|
| 16 B | −48.03 % | −51.05 % | −46.13 % |
| 32 B | −42.52 % | −43.94 % | −43.07 % |
| 48 B | −32.83 % | −37.69 % | −40.51 % |
| 64 B | −21.53 % | −26.98 % | −31.08 % |
| 128 B and above | — | — | — |

Above 64 B, 23 of 27 cells were inside their A/A floors; the four that resolved were small (`−0.78 %`, `+0.03 %`, `−0.82 %`, `−0.93 %`). So the fused path is a short-message optimisation, not a bulk one.

**Through the aws-lc wrapper.**

At 256 B through `bssl speed`, the measured gain is:

| host | delta |
|---|---:|
| Graviton3 / V1 | −5.8 % |
| Graviton4 / V2 | −6.1 % |
| Graviton5 / V3 | −8.0 % |

nebeid:
## Dispatch threshold and AEAD results

- End-to-end AEAD decrypt results at lengths reached by the current AWS-LC dispatcher.
- Definitions for A/A controls, paired deltas, uncertainty, and threshold terminology.

---

Follow-up note on the library-level numbers and on the benchmarking terminology.

**What "shipped threshold" means here**

This means current `aws-lc` behaviour as it exists today. On AArch64, `crypto/fipsmodule/modes/gcm.c` only routes to the 8x GCM decrypt kernel when `len >= 256`; below that, the call stays on the existing fallback path instead of reaching the 8x kernel.

So the table below is the end-to-end `aws-lc` AEAD decrypt/open path as a caller sees it today, not the raw kernel harness.

Negative = this PR is faster than current `aws-lc`.

| size | Graviton3 / V1 | Graviton4 / V2 | Graviton5 / V3 |
|---:|---:|---:|---:|
| 256 B | −5.8 % | −6.1 % | −8.0 % |
| 512 B | −4.4 % | −7.3 % | −6.3 % |
| 1024 B | −3.9 % | −3.8 % | −4.3 % |
| 4096 B | −0.7 % | −1.4 % | −1.5 % |

Raw `ns/call` behind that table (`pristine -> dec_opt`):

| size | V1 | V2 | V3 |
|---:|---:|---:|---:|
| 256 B | 166.8 -> 157.2 | 142.7 -> 134.0 | 119.1 -> 109.6 |
| 512 B | 221.0 -> 211.2 | 188.6 -> 174.8 | 152.9 -> 143.2 |
| 1024 B | 311.6 -> 299.5 | 261.8 -> 251.9 | 221.1 -> 211.5 |
| 4096 B | 825.3 -> 819.5 | 729.3 -> 719.2 | 629.7 -> 620.3 |

That is why the earlier summary only called out 256 B: it is the first size that the shipped `aws-lc` dispatcher actually routes into this kernel. The larger table shows the shape more clearly: real gains at reachable mid-sizes, then decay by 4 KB as wrapper and non-kernel work dominate.

**Method note in plain English**

- `A/A` means "A versus A": we compare the same code against an identical twin. This measures how much apparent difference the harness can produce from placement and run-to-run noise alone.
- `A/A floor` or `noise floor` means: if a claimed speedup is smaller than what the identical-twin comparison already produces, we do not treat that claim as resolved.
- `Paired deltas` means each run compares two variants under the same run conditions, then records the per-run difference. This is stronger than timing one variant now and the other later.
- `Median of per-process paired deltas` means we report the middle result across many paired runs, rather than the minimum. That avoids one lucky address-space layout or one unusually good run dominating the conclusion.
- `Bootstrap CI` means the uncertainty interval is estimated by resampling the observed run-to-run deltas many times. In practice: the interval comes from the measured data, not from assuming a perfect textbook distribution.
- `taskset`-pinned means each worker is pinned to a fixed CPU so scheduler movement does not blur the result.
- `Shipped threshold` means the current `aws-lc` dispatch threshold as committed today.
- `Patched threshold` means a hypothetical build where that threshold is lowered to let shorter messages reach this kernel.

nebeid:
## In-tree validation and compact-path tradeoffs

- In-tree reproduction of the current fused path and separate exact-size paths.
- Code-size and performance comparisons for t4, t4p8, and the full eight-body design.

---

Preserving the **s2n-bignum in-tree harness** results as well. The existing
benchmark comments above use the custom co-linked harness; this rerun uses the
PR's own `benchmarks/benchmark.c` (`dc71bcbfa230cdeb6db333ff4101eb23`).

Reproduced on Graviton3/V1, Graviton4/V2, and Graviton5/V3 at PR head
`29c53264`: 20 round-robin rounds per host, one executable per kernel, pinned to
CPU 3. A second independent current-kernel build was bit-identical and served
as the A/A control. Each kernel's bytes were also found at the same offset in
its linked benchmark executable.

**Current fused 1-4-block path versus the pre-fusion kernel**

Negative means the current PR is faster. `~` means below the A/A plus
0.1 ns print-resolution floor.

| size | Graviton3 / V1 | Graviton4 / V2 | Graviton5 / V3 |
|---:|---:|---:|---:|
| 16 B | -34.13% | -43.75% | -41.22% |
| 32 B | -35.63% | -37.46% | -39.53% |
| 48 B | -29.44% | -31.50% | -33.21% |
| 64 B | -22.01% | -23.58% | -26.09% |
| 128 B | -0.11%~ | +0.58% | +0.00%~ |
| 256 B | -0.16%~ | -0.18%~ | +0.00%~ |
| 512 B | -0.46% | +0.21% | +0.00%~ |
| 1024 B | -0.54% | +0.06%~ | +0.00%~ |
| 4096 B | -0.73% | +0.01%~ | -0.02%~ |

This reproduces the previous in-tree result closely. For example, the earlier
16 B figures were `-33.87% / -43.75% / -41.22%`.

I also reran the **eight separate small-size paths** experiment to
check the speed/code-size compromise against the current PR:

| size | Graviton3 / V1 | Graviton4 / V2 | Graviton5 / V3 |
|---:|---:|---:|---:|
| 16 B | -7.26% | -2.92% | -0.70% |
| 32 B | -3.44% | -5.58% | -2.87% |
| 48 B | -7.93% | -8.04% | -5.03% |
| 64 B | -14.09% | -12.11% | -4.90% |
| 128 B | -5.76% | -3.91% | -2.87% |
| 256 B and above | unresolved | unresolved | unresolved |

Those are the in-tree figures, where setup work inside the timed region
compresses kernel deltas. The co-linked raw-kernel harness also reproduced the
old experiment: the separate paths' clearest additional gains are at
64-128 B, and every host is unresolved at 256 B and above.

The exact object sizes are:

| design | `.text` |
|---|---:|
| current PR fused path | 5,960 B |
| four separate paths (16--64 B) | 7,312 B |
| five separate paths (16--64 B and 128 B) | 8,832 B |
| eight separate paths | 12,376 B |

So the separate-path design adds 6,416 bytes and makes the current kernel
`2.08x` as large, for gains confined to short messages. This quantifies the
speed, code-size, and proof-path tradeoff without selecting between the shared
and separate-path designs.

For completeness, I also measured two truncated forms. `t4` retains only the
exact-size bodies for 16, 32, 48, and 64 bytes. Its `.text` is 7,312 bytes:
1.47x the 4,968-byte pre-fusion kernel and 1,352 bytes larger than the current
PR kernel. It leaves 128 bytes on the existing path.

The performance-equivalent decrypt analogue of the
[compact encrypt experiment](https://github.com/nebeid/s2n-bignum/tree/aes-gcm-enc-fast1-4-experiment/_docs/aes-gcm-enc-fast1-4-experiment)
is `t4p8`. It retains exact-size bodies `{1,2,3,4,8}` for 16--64 and 128 bytes,
costs 8,832 bytes (1.78x), and leaves 80, 96, and 112 bytes on the existing
path. Versus the pre-fusion kernel:

| size | Graviton3 / V1 | Graviton4 / V2 | Graviton5 / V3 |
|---:|---:|---:|---:|
| 16 B | -46.86% | -47.27% | -46.27% |
| 32 B | -43.84% | -42.83% | -42.40% |
| 48 B | -43.56% | -43.48% | -44.92% |
| 64 B | -41.29% | -40.81% | -42.01% |
| 128 B | -9.72% | -9.95% | -8.61% |

The in-tree separate-body rerun above gives the incremental change relative to
the current shared path at those sizes:

| size | Graviton3 / V1 | Graviton4 / V2 | Graviton5 / V3 |
|---:|---:|---:|---:|
| 16 B | -7.26% | -2.92% | -0.70% |
| 32 B | -3.44% | -5.58% | -2.87% |
| 48 B | -7.93% | -8.04% | -5.03% |
| 64 B | -14.09% | -12.11% | -4.90% |
| 128 B | -5.76% | -3.91% | -2.87% |

Those incremental figures use the eight-body object, whose retained bodies are
the same exact-size designs. The independent `t4p8` experiment reproduced the
full object's 128-byte gain to 0.08 percentage points against a
placement-matched control, while 80/96/112 bytes and 256 bytes and above were
unchanged. Report and generator:
[`fused-t4p8.md`](https://github.com/nebeid/s2n-bignum/blob/aes-gcm-fused-wip/_docs/fused-t4p8.md)
and
[`gen_set.py`](https://github.com/nebeid/s2n-bignum/blob/aes-gcm-fused-wip/_docs/fused-t4p8/gen_set.py).

arm/proofs/aesv8_gcm_8x_dec_256_wb.ml
```diff
@@ -0,0 +1,16895 @@
...
+    (word:num->(64)word) (128 * (nblk:num)) /\
+    (read:(armstate,(64)word)component->armstate->(64)word)
+    (X2:(armstate,(64)word)component)
+    (s:armstate) =
+    (out_p:(64)word) /\
+    (read:(armstate,(64)word)component->armstate->(64)word)
+    (X3:(armstate,(64)word)component)
+    (s:armstate) =
+    (xi_p:(64)word) /\
+    (rea
```
manastasova: nit: `s /\` alignment -- of previous line

PR #446, opened 2026-08-25
Fix x86 Curve25519 Mach-O table references

justsmth:
## Summary

- Apply `S2N_BN_SYMBOL` to the remaining x86 X25519 and Edwards25519 table references and regenerate the AT&T assembly.
- Add ARM and x86 Mach-O CI checks to prevent symbol-name mismatches.

This fixes issues introduced when the tables moved to read-only data in #242, #413, and #414. Found while investigating aws/aws-lc#3408.

## Testing

- Verified Intel and AT&T objects match and ELF objects are byte-for-byte unchanged for native and Windows ABI modes.
- Confirmed all affected x86 functions link on Mach-O and all 690 ARM/x86 library objects have no undefined symbols.
- Ran 100 `curve25519_x25519base_alt` cases against the bundled TweetNaCl reference under Rosetta.


jargh (approved):
All looks good, thank you! The fix to the .S files is right and the additional checks in CI should be useful for catching related issues in the future. I'll merge.

PR #449, opened 2026-09-01
Move loop-based memory-safety tactics into common/consttime.ml

sgmenda:
Deduplicate MEMSAFE_ARITH_TAC, CONTAINED_ASM_TAC and DISCHARGE_MEMSAFE_ASM_TAC, previously copied in both arm and x86 mlkem_rej_uniform_VARIABLE_TIME.ml proofs, into common/consttime.ml so upcoming loop-based memory-safety proofs can share them.


arm/proofs/mlkem_rej_uniform_VARIABLE_TIME.ml
manastasova, line 1677: Would it be worth moving `DISCHARGE_MEMSAFE_TAC` as well? (also in other files)

common/consttime.ml
```diff
@@ -1020,4 +1020,70 @@ let DISCHARGE_SAFETY_PROPERTY_TAC =
...
+   contained_modulo bounds with symbolic loop variables.
+*)
+let DISCHARGE_MEMSAFE_ASM_TAC:tactic =
+  SAFE_META_EXISTS_TAC allowed_vars_e THEN
+  CONJ_TAC THENL [ EXISTS_E2_TAC allowed_vars_e; ALL_TAC ] THEN
+  REWRITE_TAC[MEMACCESS_INBOUNDS_APPEND] THEN
+  CONJ_TAC THENL
+   [REWRITE_TAC[memaccess_inbounds; ALL; EX; FST; SND] THEN
+    REPEAT CONJ_TAC THEN
+    TRY(REPEAT ((DISJ1_TAC THEN CONTAINED_ASM_TAC) ORELSE DISJ2_TAC ORELSE
```
manastasova, line 1086: nit: Is `TRY` needed?

jakemas:
Looks good to me! One of the only things to look out for when ripping stuff from an x86/Aarch64 proof file to the generic common is that sometimes there may be x86/arm specific tactics within the proof. I'll keep an eye on CI.

jakemas (approved):

common/consttime.ml
sgmenda: Good question. I tested it. It is not needed for the `x86` proof which always succeeds. But it is needed for the `arm` proof which can return a `NO_TAC`.

I did not dig into why the `arm` proof can return `NO_TAC`.

arm/proofs/mlkem_rej_uniform_VARIABLE_TIME.ml
sgmenda: Yep, same pattern, but that's a larger change that touches `mldsa` too. I stuck to what I need for my XTS safety proofs.

sgmenda:
@jakemas good callout. none of the stuff I moved to common seemed to be arch-specific. I've been relying the proofs passing (locally and on CI) to confirm such intuition.

manastasova (approved):

PR #451, opened 2026-09-08
AES-XTS encrypt: prove the len<0x20 memory-safety lemma

sgmenda:
Prove `AES_XTS_ENCRYPT_LT_2BLOCK_MEMSAFE`: for `AES-256-XTS` encrypt of a single full block plus optional tail (16 <= len < 32), every memory access stays in-bounds. Memory-safety only (in-bounds event trace), no `CHEAT`ing.

The readable region includes `ctxt_p`: cipherstealing reads back the previously-written ciphertext, which the caller owns. Rationale and an address-layout diagram are in the comment at the lemma. The `MAYCHANGE` frame is declared narrowly up front with `ENSURES_FRAME_SUBSUMED` at each segment, matching the idiom already used elsewhere in this file. (h/t @manastasova) 

This is the first per-length-class lemma; the general `AES256_XTS_ENCRYPT_MEMSAFE` will dispatch on len over these (mirroring `AES256_XTS_ENCRYPT_CORRECT`). Follow-up PRs will add the remanining length classes, constant-time, and decrypt.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.


arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,456 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+(*                                                                           *)
+(* These fill small gaps in the shared subsumed/frame machinery for the      *)
+(* variable-length store patterns that arise here. They are kept file-local  *)
+(* (mirroring arm/proofs/mlkem_rej_uniform_VARIABLE_TIME.ml); the two        *)
+(* MAYCHANGE_BYTES{8,128}_SUBSUMED_BYTES lemmas are candidates for later     *)
+(* upstreaming to common/relational.ml in a separate change.                 *)
+(* ------------------------------------------------------------------------- *)
+
+(* Transitivity of `subsumed` (no exported equivalent in common/relational). *)
+let SUBSUMED_TRANS = prove
```
manastasova: I spent a while understanding why these are needed. I see the point — with a variable output length and cipher stealing, the declared frame is `bytes(ctxt_p, val len)` while the actual stores are `bytes128 ctxt_p` and single `bytes8` at `ctxt_p+i`, so something has to bridge the two.
But I'd suggest simplifying if possible.
Claude's suggestion: simplify by declaring the narrow frame up front with ENSURES_FRAME_SUBSUMED, rather than repairing an accumulated one afterwards. This file already does exactly that at line 2080 — for the very same two cipher-stealing strb stores. With the frame written out literally, all four lemmas here can be skipped: SUBSUMED_APPLY has no state-level goal left to bridge, SUBSUMED_TRANS has no intermediate frame to insert, and both MAYCHANGE_BYTES*_SUBSUMED_BYTES are handled by the existing SUBSUMED_SEQ / ABBREV_TAC vallen / SUBSUMED_MAYCHANGE_TAC sequence that line 2080 uses. ENSURES_FINAL_STATE_TAC then closes the frame goal by itself, so the trailing FRAME_SUBSUMED_TAC / LOOP_FRAME_TAC can go too.

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,456 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+  MATCH_MP_TAC SUBSUMED_ASSIGNS_SUBCOMPONENTS THEN MATCH_MP_TAC SUBSUMED_TRANS THEN
+  EXISTS_TAC `ASSIGNS (bytes (b:int64,1))` THEN CONJ_TAC THENL
+  [MATCH_ACCEPT_TAC SUBSUMED_ASSIGNS_SUBCOMPONENT;
+   MATCH_MP_TAC SUBSUMED_ASSIGNS_BYTES THEN ASM_REWRITE_TAC[]]);;
+
+(* Term surgery over the `,,`-sequenced MAYCHANGE frame. Names are prefixed  *)
+(* `xts_ms_` since HOL Light has a single flat OCaml namespace shared across *)
+(* the whole build; these generic-looking helpers must not collide with (or  *)
+(* be shadowed by) same-named bindings in other proof files.                 *)
+let xts_ms_seqop =
```
manastasova: Same here. I haven't come across this kind of symbol fixing in the proofs.
Claude: Same suggestion as on SUBSUMED_TRANS: if the frame is declared up front with ENSURES_FRAME_SUBSUMED, there's no ,,-chain to take apart and rebuild, so this whole group goes with it.

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,456 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+    (EXISTS_TAC prefix THEN CONJ_TAC THENL
+     [ ASM_REWRITE_TAC[APPEND];
+       REWRITE_TAC[memaccess_inbounds; ALL; EX] THEN
+       REWRITE_TAC[contained; DIMINDEX_64] THEN REPEAT CONJ_TAC THEN
+       FIRST [ REWRITE_TAC[CONTAINED_MODULO_REFL] THEN DISJ1_TAC THEN ASM_ARITH_TAC;
+               REPEAT ((DISJ1_TAC THEN CONTAINED_TAC) ORELSE DISJ2_TAC ORELSE
+                       CONTAINED_TAC) ] ]) (asl,w);;
+
+(* Recognisers for ctxt_p-based memory components in a frame. *)
+let xts_ms_is_ctxt_bytes8 c =
```
manastasova: Same as above

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,456 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
       fst AES256_XTS_ENCRYPT_EXEC] AES256_XTS_ENCRYPT_CORRECT)
    `[X19; X20; X21; X22; D8; D9; D10; D11; D12; D13; D14; D15]` 96
   );;
+
+(* ========================================================================= *)
+(* AES-256-XTS encryption: memory safety.                                    *)
+(*                                                                           *)
+(* Proves that all memory accesses performed by aes256_xts_encrypt stay      *)
```
manastasova: nit: the prologue/epilogue stack accesses aren't proven yet — adding "the body of" before aes256_xts_encrypt would make that sentence a bit more accurate.

manastasova (commented):
The goal definition looks good to me - it matches `AES_XTS_ENCRYPT_LT_2BLOCK_CORRECT`, and the readable/writable lists look right.
I haven't gone through every lemma/tactic, so please don't take this as a blocker. I was wondering if you could try avoiding the four `SUBSUMED_*` lemmas and the `xts_ms_*` surgery group?

sgmenda:
Thanks, @manastasova for the tip.

Per your suggestion, I narrowed the frame up front with `ENSURES_FRAME_SUBSUMED`, which let me drop the four `SUBSUMED_*`/`MAYCHANGE_BYTES*` lemmas and the `xts_ms_*` surgery.

manastasova (approved):

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,339 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+(* trace). Constant-time (the `exists f_events` framing) is deliberately     *)
+(* out of scope here; this is the memory-safety-only variant, so the         *)
+(* theorem carries the `_MEMSAFE` suffix (cf. MLKEM_REJ_UNIFORM_MEMSAFE).    *)
+(*                                                                           *)
+(* Note ctxt_p appears as a *readable* region as well as writable: AES-XTS   *)
+(* ciphertext stealing reads back the previously-written ciphertext block    *)
+(* (see the comment on the read-back at the cipher-stealing tail below).     *)
+(* ========================================================================= *)
+
+needs "arm/proofs/consttime.ml";;
```
nebeid, line 5630: Rebase needed due to #452's generalization for 32-bit backends

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,339 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+(* aes256_xts_encrypt stay within the permitted buffers:                     *)
+(*   readable: plaintext [ptxt_p,len],                                       *)
+(*             ciphertext [ctxt_p,len],                                      *)
+(*             iv [iv_p,16],                                                 *)
+(*             key schedules [key1_p,244] and [key2_p,244]                   *)
+(*   writable: ciphertext [ctxt_p,len]                                       *)
+(*                                                                           *)
+(* Mechanics follow arm/tutorial/safety.ml (memaccess_inbounds / event       *)
+(* trace). Constant-time (the `exists f_events` framing) is deliberately     *)
+(* out of scope here; this is the memory-safety-only variant, so the         *)
```
nebeid: **On "constant-time ... deliberately out of scope": there is already a CT proof of this exact byte range.** Branch `aes-xts-safety` (tip `2e362c73`, unmerged) has `CIPHER_STEALING_ENC_SAFE` at `aes_xts_encrypt.ml:2547` over entry `pc + 0x9e0` → exit `pc + 0xa8c` — the same region as SEG2 here — in the full `exists f_events` form, and with hypothesis `val len >= 16 /\ val len <= 2 EXP 24` rather than `val len < 0x20`, i.e. all length classes, not one. 376 lines for the whole encrypt tail vs. 336 here for one class, memsafe-only.

So the staging question isn't really "is memsafe-only an acceptable first step" but "why not land the `f_events` form directly." Caveats, stated plainly: that branch is ~7 months old, I have **not** re-run it, and it needs one unmerged library hook (`?abbrevs_unfold_before_f_events_tac` on `DISCHARGE_SAFETY_PROPERTY_TAC`, absent from `upstream/main:common/consttime.ml`). Which of the two rebases more cheaply is unknown. But the two will conflict — both insert into this file.

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,339 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+(* theorem carries the `_MEMSAFE` suffix (cf. MLKEM_REJ_UNIFORM_MEMSAFE).    *)
```
nebeid: **Naming: `_MEMSAFE` is right here, but flag the export now.** For an internal per-length-class lemma the `_MEMSAFE` suffix matches precedent exactly — `MLKEM_REJ_UNIFORM_MEMSAFE` at `arm/proofs/mlkem_rej_uniform_VARIABLE_TIME.ml:1693`. No change needed.

The thing to pin down is the *exported* theorem in the follow-up: `tools/count-proofs.sh:22` greps literally `'SUBROUTINE_CORRECT : thm'` / `'SUBROUTINE_SAFE : thm'`, so a `..._SUBROUTINE_MEMSAFE` export would be invisible to the `specifications.txt` diff gate. Same file's precedent: `MLKEM_REJ_UNIFORM_SUBROUTINE_SAFE` (`:3024`), listed at `arm/proofs/specifications.txt:377`. (Nothing safety-related for AES-XTS is in `specifications.txt` on `upstream/main` yet, so this PR itself needs no `specifications.txt` edit.)

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,339 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+(* mirroring arm/proofs/mlkem_rej_uniform_VARIABLE_TIME.ml. The MAYCHANGE    *)
+(* frame is narrowed up front with ENSURES_FRAME_SUBSUMED at each segment    *)
+(* (as elsewhere in this file), so no bespoke frame-widening machinery is    *)
+(* needed here.                                                              *)
+(* ------------------------------------------------------------------------- *)
+
+(* Discharge `exists e2. read events s = APPEND e2 e /\ memaccess_inbounds`  *)
+(* for a CONCRETE event trace (events = CONS ... e), as arises at a          *)
+(* straight-line final state. The prefix e2 is read off the events equation. *)
+let SAFETY_EXISTS_TAC : tactic =
```
nebeid: **`SAFETY_EXISTS_TAC` duplicates a library tactic that already existed at this PR's own merge base.** `DISCHARGE_MEMSAFE_ASM_TAC` is at `21f6fa62:common/consttime.ml:1079` — i.e. available without a rebase — and its body is the same composition (`REWRITE_TAC[MEMACCESS_INBOUNDS_APPEND]`, then `memaccess_inbounds; ALL; EX; FST; SND` + `CONTAINED_ASM_TAC`/`DISJ` sweep, then accept the tail).

Two fragilities in the bespoke version, both failure modes rather than unsoundness:
- the `find` over `asl` takes the *first* assumption whose LHS mentions `events`, which also matches the precondition `read events s0 = e`; on that match `split_at_e` returns `([], e)` and `hd hs` raises.
- `ignore tl` — the tail is never checked to be `e`, so nothing verifies the prefix is the whole delta.

Suggest deleting it and using the library tactic.

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,339 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+(*                                                                           *)
+(*     ptxt_p           ptxt_p+16             ptxt_p+len                     *)
+(*       +-----------------+-----------------------+                         *)
+(*       |  full block 0   |    tail (tail_len B)  |   plaintext buffer      *)
+(*       +-----------------+-----------------------+                         *)
+(*                          ^ ldrb ptxt_p+16+i                               *)
+(*                                                                           *)
+(* Every access offset is < len, so all lie within the named regions.        *)
+(* ------------------------------------------------------------------------- *)
+let AES_XTS_ENCRYPT_LT_2BLOCK_MEMSAFE = prove(
```
nebeid, line 5713: Its correctness twin `AES_XTS_ENCRYPT_LT_2BLOCK_CORRECT` (`:2547`) is `time prove`, as is every other simulation-level theorem in this file: `:1509`, `:2722`, `:2891`, `:3064`. The only two that aren't — `AES256_XTS_ENCRYPT_CORRECT` (`:3246`) and `AES_XTS_ENCRYPT_SUBROUTINE_CORRECT` (`:5563`) — are the ones that just dispatch and wrap, with no `ARM_STEPS_TAC`. This lemma is on the simulation side of that line.

With four more length classes plus decrypt queued, that timing line is the only per-theorem cost signal — CI reports just the 4h17m aggregate.

```suggestion
let AES_XTS_ENCRYPT_LT_2BLOCK_MEMSAFE = time prove(
```

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,339 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+  `!ptxt_p ctxt_p len key1_p key2_p iv_p
+    pt_in iv
+    k1_0 k1_1 k1_2 k1_3 k1_4 k1_5 k1_6 k1_7 k1_8 k1_9 k1_10 k1_11 k1_12 k1_13 k1_14
+    k2_0 k2_1 k2_2 k2_3 k2_4 k2_5 k2_6 k2_7 k2_8 k2_9 k2_10 k2_11 k2_12 k2_13 k2_14
+    pc e.
+    PAIRWISE nonoverlapping
```
nebeid: `iv_p` is claimed readable in the postcondition (`[... iv_p,16 ...]`, below) but is absent from this `PAIRWISE nonoverlapping` list. So as stated the lemma permits `iv_p` to alias the code region, `key1_p`, `key2_p` or `ctxt_p` while still asserting its reads are in bounds.

Not introduced by this PR — inherited verbatim from `AES_XTS_ENCRYPT_LT_2BLOCK_CORRECT` (`upstream/main:arm/proofs/aes_xts_encrypt.ml:2552-2557`) — so I'm not asking for a change here. But what will the subroutine-level spec require? Worth deciding now rather than at export time.

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,339 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+           read(memory :> bytes128 iv_p) s = iv /\
+           set_key_schedule s key1_p k1_0 k1_1 k1_2 k1_3 k1_4 k1_5 k1_6 k1_7 k1_8 k1_9 k1_10 k1_11 k1_12 k1_13 k1_14 /\
+           set_key_schedule s key2_p k2_0 k2_1 k2_2 k2_3 k2_4 k2_5 k2_6 k2_7 k2_8 k2_9 k2_10 k2_11 k2_12 k2_13 k2_14 /\
+           read events s = e
+      )
+      // postcondition
+      (\s. read PC s = word (pc + 0xa8c) /\
+           (exists e2.
+                read events s = APPEND e2 e /\
+                memaccess_inbounds e2
```
nebeid: These region lists are hand-typed in four places (here, plus the SEG1 intermediate assertion and both loop invariants) and checked by nothing. `aes_xts_encrypt` is already in `arm/proofs/subroutine_signatures.ml:25`, so the eventual export can get its lists from `mk_safety_spec (assoc "aes_xts_encrypt" subroutine_signatures) ...` (`arm/proofs/consttime.ml:18`; usage pattern at `arm/proofs/bignum_ge.ml:368-374`) instead. The generator doesn't apply to this intermediate `pc+28 → pc+0xa8c` lemma directly, but it's worth checking the hand-typed lists against what it will produce, so the export doesn't drift from the lemma it's built on.

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,339 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+        ARM_STEPS_TAC AES256_XTS_ENCRYPT_EXEC (11--33) THEN
+        ENSURES_FINAL_STATE_TAC THEN ASM_REWRITE_TAC[] THEN
+        (EXISTS_TAC
+             `APPEND [EventStore (ctxt_p:int64,16); EventLoad (ctxt_p:int64,16); EventJump (word(pc+2568),word(pc+2572))] e_acc` THEN
+           CONJ_TAC THENL
+           [ REWRITE_TAC[APPEND];
+             REWRITE_TAC[MEMACCESS_INBOUNDS_APPEND] THEN CONJ_TAC THENL
+             [ (REWRITE_TAC[memaccess_inbounds; ALL; EX; FST; SND] THEN REPEAT CONJ_TAC THEN
+                REPEAT ((DISJ1_TAC THEN CONTAINED_ASM_TAC) ORELSE DISJ2_TAC ORELSE CONTAINED_ASM_TAC));
+               FIRST_ASSUM ACCEPT_TAC ]])]]]);;
```
nebeid: This closing block is `DISCHARGE_MEMSAFE_ASM_TAC`'s body inlined (`21f6fa62:common/consttime.ml:1079-1090`) with the `SAFE_META_EXISTS_TAC`/`EXISTS_E2_TAC` front replaced by an `EXISTS_TAC` of a literal `[EventStore; EventLoad; EventJump]` list. Same point as on `SAFETY_EXISTS_TAC` above: the library tactic gets you here without hand-written event literals, which is what makes these leaves survive an instruction-schedule change.

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,339 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+        `val (len:int64) >= 0x10 ==> val len < 0x20 ==>
+         ival (len:int64) - &0x10 = ival (word_sub len (word 0x10))`) THEN
+      ASM_REWRITE_TAC[] THEN ARITH_TAC; ALL_TAC] THEN
+    ASM_REWRITE_TAC[] THEN DISCH_TAC THEN
+    POP_ASSUM(fun th -> RULE_ASSUM_TAC(REWRITE_RULE[th])) THEN
+    (* One full block, so len_full_blocks = 16. *)
+    SUBGOAL_THEN `val (word_and (len:int64) (word 18446744073709551600)) = 16` ASSUME_TAC THENL
+    [ MP_TAC(BITBLAST_RULE
+        `val (len:int64) >= 16 ==> val len < 32 ==>
+         val (word_and len (word 18446744073709551600)) = 16`) THEN
```
nebeid: Worth a one-line comment stating the bound out loud: the in-bounds obligation for the byte loop needs `16 + i < val len`, which follows from the `BITBLAST_RULE` fact `val len = 16 + val(word_and len (word 15))` together with PADOWN's body hypothesis `0 <= i /\ i < val(word_and len (word 15))`. It does hold — I traced it — but it's the one step a reader has to reconstruct, and it's the step that changes when the `val len < 0x20` hypothesis is relaxed for the other length classes.

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,339 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+    let tl,hs = split_at_e rhs_t in ignore tl;
+    let prefix = mk_list(hs, type_of(hd hs)) in
+    (EXISTS_TAC prefix THEN CONJ_TAC THENL
+     [ ASM_REWRITE_TAC[APPEND];
+       REWRITE_TAC[memaccess_inbounds; ALL; EX] THEN
+       REWRITE_TAC[contained; DIMINDEX_64] THEN REPEAT CONJ_TAC THEN
+       FIRST [ REWRITE_TAC[CONTAINED_MODULO_REFL] THEN DISJ1_TAC THEN ASM_ARITH_TAC;
+               REPEAT ((DISJ1_TAC THEN CONTAINED_TAC) ORELSE DISJ2_TAC ORELSE
+                       CONTAINED_TAC) ] ]) (asl,w);;
+
```
nebeid, line 5663: `DISCHARGE_MEMSAFE_ASM_TAC` already does this, from your own cbc763aa ("Hoist loop-based memory-safety tactics into common/consttime.ml", 2026-09-01) — which is also the commit that emptied the `mlkem_rej_uniform_VARIABLE_TIME.ml` precedent cited at 5633-5634. No rebase needed to reach it: `common/consttime.ml:1079` on this head, in scope via `arm/proofs/consttime.ml:10`.

Two bugs in the local copy, both failure modes rather than unsoundness: the `find` can match the precondition `read events s0 = e` (then `hd hs` raises), and `ignore tl` never checks the tail is `e`.

```suggestion
```

Paired with the swap at 5802. Untested — I have not run HOL Light.

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,339 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+    POP_ASSUM(fun th -> RULE_ASSUM_TAC(REWRITE_RULE[th])) THEN
+    (* One full block, so len_full_blocks = 16. *)
+    SUBGOAL_THEN `val (word_and (len:int64) (word 18446744073709551600)) = 16` ASSUME_TAC THENL
+    [ MP_TAC(BITBLAST_RULE
+        `val (len:int64) >= 16 ==> val len < 32 ==>
+         val (word_and len (word 18446744073709551600)) = 16`) THEN
+      ASM_REWRITE_TAC[]; ALL_TAC] THEN
+    ARM_STEPS_TAC AES256_XTS_ENCRYPT_EXEC (3--118) THEN
+    ENSURES_FINAL_STATE_TAC THEN
+    REPEAT CONJ_TAC THEN TRY(FIRST_X_ASSUM ACCEPT_TAC) THEN SAFETY_EXISTS_TAC;
```
nebeid, line 5802:
```suggestion
    REPEAT CONJ_TAC THEN TRY(FIRST_X_ASSUM ACCEPT_TAC) THEN DISCHARGE_MEMSAFE_ASM_TAC;
```

Untested: `DISCHARGE_MEMSAFE_ASM_TAC` opens with `SAFE_META_EXISTS_TAC allowed_vars_e` (metavariable) where the local tactic does an explicit `EXISTS_TAC`, so `allowed_vars_e` may need seeding first. Worth a trial before taking it.

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,339 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+    pt_in iv
+    k1_0 k1_1 k1_2 k1_3 k1_4 k1_5 k1_6 k1_7 k1_8 k1_9 k1_10 k1_11 k1_12 k1_13 k1_14
+    k2_0 k2_1 k2_2 k2_3 k2_4 k2_5 k2_6 k2_7 k2_8 k2_9 k2_10 k2_11 k2_12 k2_13 k2_14
+    pc e.
+    PAIRWISE nonoverlapping
+    [(word pc, LENGTH aes256_xts_encrypt_mc);
+    (ptxt_p, val len);
+    (ctxt_p, val len);
+    (key1_p, 244);
+    (key2_p, 244)] /\
```
nebeid: `iv_p` is claimed readable in the postcondition (`[... iv_p,16 ...]`) but is absent here, so as stated the lemma permits `iv_p` to alias the code region, `key1_p`, `key2_p` or `ctxt_p` while still asserting its reads are in bounds.

Inherited from `AES_XTS_ENCRYPT_LT_2BLOCK_CORRECT` (`upstream/main:arm/proofs/aes_xts_encrypt.ml:2552-2557`), not introduced here. Suggestion below if you want it closed at this level; otherwise `AES256_XTS_ENCRYPT_MEMSAFE` has to state it, and deciding now beats four more classes inheriting the gap.

```suggestion
    PAIRWISE nonoverlapping
    [(word pc, LENGTH aes256_xts_encrypt_mc);
    (ptxt_p, val len);
    (ctxt_p, val len);
    (iv_p, 16);
    (key1_p, 244);
    (key2_p, 244)] /\
```

arm/proofs/aes_xts_encrypt.ml
nebeid, line 5943: Same point as the `SAFETY_EXISTS_TAC` thread: `DISCHARGE_MEMSAFE_ASM_TAC`'s body open-coded (`common/consttime.ml:1079-1090`, your cbc763aa), with `SAFE_META_EXISTS_TAC`/`EXISTS_E2_TAC` swapped for an `EXISTS_TAC` of a literal event list. The library tactic avoids the literals, which is what lets these leaves survive an instruction-schedule change — otherwise each remaining length class needs its own list.

```suggestion
        DISCHARGE_MEMSAFE_ASM_TAC]]]);;
```

Untested, and this one needs a careful look: the suggestion keeps the `]]]);;` outer closers and drops the `(`/`[`/`[` opened on the replaced lines. Check the balance before taking it.

arm/proofs/aes_xts_encrypt.ml
```diff
@@ -5605,3 +5605,339 @@ let AES_XTS_ENCRYPT_SUBROUTINE_CORRECT = prove(
...
+            [ REWRITE_TAC[]; REWRITE_TAC[GSYM memaccess_inbounds] THEN FIRST_X_ASSUM ACCEPT_TAC]]];
+
+        (* (3) loop body: invariant@(i+1) @pc+2548 -> invariant@i @pc+2568. *)
+        (* In order the body does: ldrb ctxt_p+i (the ciphertext *)
+        (* read-back), ldrb ptxt_p+16+i, strb ctxt_p+16+i, strb ctxt_p+i. *)
+        (* The frame is narrowed to the two per-byte stores up front, and *)
+        (* ARM_STEPS derives their code non-overlap from the PAIRWISE *)
+        (* nonoverlapping hypothesis. *)
+        REPEAT STRIP_TAC THEN
+        SUBGOAL_THEN `val (len:int64) = 16 + val (word_and len (word 15))` ASSUME_TAC THENL
```
nebeid, line 5853: Optional nit. The `16 + i < val len` obligation is already an explicit subgoal at 5858, so nothing is hidden. What's missing is why the equation on this line holds — it is exactly what `val len < 0x20` buys, and it's the step each remaining length class has to re-derive.

```suggestion
        (* 16 <= val len < 0x20 means exactly one full block, so the tail is *)
        (* val len - 16. Each length class re-derives its own version.       *)
        SUBGOAL_THEN `val (len:int64) = 16 + val (word_and len (word 15))` ASSUME_TAC THENL
```

nebeid (commented):
No soundness objection — the lemma statement and the `val len < 0x20` bound check out.

Two substantive items: the **rebase**, and the file-local `SAFETY_EXISTS_TAC`, which re-implements `DISCHARGE_MEMSAFE_ASM_TAC` from your own cbc763aa hoist — three sites (the definition, the use at 5802, and the open-coded copy at 5936-5943). Then a `time prove` nit and one optional comment request.

Suggestions attached where they apply, but I have not run HOL Light — treat the tactic ones as proposals to try.

PR #452, opened 2026-09-10
Generalize ISA proof infrastructure for 32-bit backends

hanno-becker:
## Goal

Make it easier to add new ISA backends, especially 32-bit backends, by reusing more of the existing proof code.

## Changes

- Make memory and address reasoning work with different address widths.
- Make events and safety proofs use the backend's address width.
- Share execution, code-loading, instruction-fetch, and alignment support.
- Support both ELF32 and ELF64 object files.
- Move general word-list and memory helpers into common code.
- Add common helpers for relational and arithmetic proofs.

A new backend still provides its state, decoder, instruction semantics, relocation handling, and ABI rules.

xisting ARM and x86 proofs keep their current interfaces and 64-bit behavior.


common/int_linear.ml
```diff
@@ -0,0 +1,159 @@
+(*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT-0
+ *)
+
+(* ========================================================================= *)
```
hanno-becker, line 6: TODO: Check if this is still used in the NTT proofs later on.

jargh (approved):
All great, thank you! I will merge this now.

PR #453, opened 2026-09-13
Arm bti support

manastasova:
x86 supports Intel CET end to end: `_CET_ENDBR` emits `ENDBR64` at each entry point, `decode.ml` decodes it, `x86.ml` gives it no-op semantics, and `define_trimmed` covers the `-DNO_IBT` build. Arm has had no counterpart. This adds the equivalent, with no `.S`, proof or `specifications.txt` changes.
- `AARCH64_VALID_CALL_TARGET` → `hint #34` (BTI c), on by default (hint space, so a NOP pre-Armv8.5-A), suppressed by `-DNO_IBT`. Named to match AWS-LC's macro, and skipped if that header already defined it.
- The required `.note.gnu.property` (ELF only). This is the main motivation: the property uses `GNU_PROPERTY_AARCH64_FEATURE_1_AND` semantics, so one object without the note silently disables BTI **program-wide** — linking s2n-bignum should not be able to turn off a consumer's BTI.
- `arm_BTI`, the identity state relation (own constant, like `x86_ENDBR64`), plus a decode clause for `0xd503245f` and an Arm `define_trimmed`.

Verified: a probe assembled with `-Iinclude` emits `d503245f bti c` at offset 0 plus the property note, and emits neither under `-DNO_IBT`; `arm/proofs/base.ml` loads clean and `ARM_MK_EXEC_RULE` gives `|- !s pc. aligned_bytes_loaded s (word pc) bti_mc ==> arm_decode s (word pc) arm_BTI` (also confirming nothing shadows the encoding); the NOP clause is unaffected; and `bti_mc_trimmed = noibt_mc` is provable against the same source assembled independently with `-DNO_IBT`.

Adopting the marker in the `.S` files and adding the BTI variants to `specifications.txt` (the checklist notes *"For ARM: 1 variant per function (BTI variants may come in the future)"*) is left to follow-ups.


include/_internal_s2n_bignum_arm.h
```diff
@@ -34,6 +34,37 @@
...
+#   endif
+#   define AARCH64_VALID_CALL_TARGET
+#elif !defined(AARCH64_VALID_CALL_TARGET)
+#   define AARCH64_VALID_CALL_TARGET hint #34 /* BTI c */
+#   ifndef __APPLE__
+        .pushsection .note.gnu.property, "a"
+        .balign 8
+        .long 4
+        .long 0x10
+        .long 0x5                 /* NT_GNU_PROPERTY_TYPE_0 */
```
sgmenda: explain magic `.long`s
```suggestion
        .long 4                   /* n_namesz: sizeof "GNU\0" */
        .long 0x10                /* n_descsz: 16 bytes of property data */
        .long 0x5                 /* n_type: NT_GNU_PROPERTY_TYPE_0 */
```

include/_internal_s2n_bignum_arm.h
```diff
@@ -34,6 +34,37 @@
...
+        .asciz "GNU"
+        .long 0xc0000000          /* GNU_PROPERTY_AARCH64_FEATURE_1_AND */
+        .long 4
+        .long 1                   /* GNU_PROPERTY_AARCH64_FEATURE_1_BTI */
+        .long 0
```
sgmenda: explain more magic `.long`s
```suggestion
        .long 0xc0000000          /* pr_type: GNU_PROPERTY_AARCH64_FEATURE_1_AND */
        .long 4                   /* pr_datasz: 4 bytes */
        .long 1                   /* pr_data: GNU_PROPERTY_AARCH64_FEATURE_1_BTI */
        .long 0                   /* pad to 8-byte alignment */
```

sgmenda (approved):
lgtm with nits

jargh:
Looks good to me too. There's a bit of asymmetry with x86 but that seems inevitable since there's nothing on ARM exactly analogous to the cet.h header file for x86. I'll merge :-)

PR #458, opened 2026-09-17
Refactor Arm/x86 cosimulation infrastructure

hanno-becker:
* ~Builds on #452~

This PR refactors the cosimulation infrastructure. Its primary goal is to enable custom cosimulation backends if the ISA under test is not the host architecture.  The ML-side cosimulation driver now communicates to a persistent cosimulation "backend" via a simple stdin/stdout protocol.  Please see the individual commit messages for details.


common/cosim.ml
```diff
@@ -0,0 +1,285 @@
...
+ * state words. OK returns the resulting state. TRAP reports an execution
+ * trap, while ERROR reports a malformed request or executor failure.
+ * The ellipses above abbreviate state words; they are not sent literally.
+ *)
+
+type cosim_response =
+  | Cosim_ok of string list
+  | Cosim_trap of string list;;
+
+let cosim_process_exit status = Unix.\_exit status;;
```
aqjune-aws: 🔴 **Critical**

`Unix.\_exit` contains a literal backslash before `_exit`, which is not valid OCaml syntax and will cause a lexer/parse error when `simulator.native` is compiled (confirmed with `cat -A`: the line is `Unix.\_exit status`). The intended function is `Unix._exit`. Since this file is a prerequisite of `proofs/simulator.native` and is loaded via `needs "common/cosim.ml"`, the entire cosimulation build (including the new CI jobs) will fail to compile. Fix by removing the backslash: `let cosim_process_exit status = Unix._exit status;;`.

---

arm/Makefile
```diff
@@ -481,7 +481,7 @@ TUTORIAL_PROOF_LOGS = $(TUTORIAL_PROOFS:.ml=.correct)
 
 # Build precompiled native binaries of HOL Light proofs
 
-proofs/simulator.native: proofs/simulator.ml $(BASE_ML) proofs/simulator_iclasses.ml ; ../tools/build-proof.sh proofs/simulator.ml "$(HOLLIGHT)" "$@"
+proofs/simulator.native: proofs/simulator.ml ../common/cosim.ml $(BASE_ML) proofs/simulator_iclasses.ml ; ../tools/build-proof.sh proofs/simulator.ml "$(HOLLIGHT)" "$@"
```
aqjune-aws: 🟡 **Suggestion**

`proofs/simulator.ml` now also `needs "common/sematest.ml"`, but only `../common/cosim.ml` was added to the prerequisites here. Edits to `common/sematest.ml` (which defines the campaign duration/case/seed controls) won't trigger a rebuild of `simulator.native`. Add `../common/sematest.ml` to the prerequisite list as well (same applies to `x86/Makefile:498`).

---

common/cosim.ml
```diff
@@ -0,0 +1,285 @@
...
+
+let cosim_words s =
+  List.filter (fun x -> x <> "") (String.split_on_char ' ' s);;
+
+let cosim_getenv name =
+  try Some (Sys.getenv name) with Not_found -> None;;
+
+let cosim_timeout () =
+  match cosim_getenv "S2N_BIGNUM_EXECUTOR_TIMEOUT" with
+  | None -> 10.0
```
aqjune-aws: 🟡 **Suggestion**

The default per-response timeout is 10s. In the new cross-cosimulation path, each RUN request causes the persistent script to invoke the legacy `armsimulate`/`x86simulate`, which recompiles/reassembles/links the harness and runs it (under QEMU for the cross jobs) on every case. If a single case's compile+link+emulated-run ever exceeds 10s on a slow CI runner, `cosim_read_line` will `failwith "timed out"` and abort the whole campaign as a spurious failure. The new CI jobs (`.github/workflows/ci.yml`) do not set `S2N_BIGNUM_EXECUTOR_TIMEOUT`. Consider raising the default or setting the env var explicitly in the cross-cosimulation CI jobs to reduce flakiness risk.

---

aqjune-aws (approved):
Looks good. It seems this also fixed a tiny timing bug in the previous cosimulator; Before this PR, the Arm sematest was running for 1 hr even if the timeout was set to 2400 seconds.

The number of instances cross-platform sematest is running seems a bit small (100) but it is fine since it is configurable from the Github Workflows file.

PR #455, opened 2026-09-16
Add partial RISCV32-IM model

hanno-becker:
* ~Builds on #452~

This PR adds a partial RV32-IM model sufficent to establish correctness of the RV32-IM backend in mldsa-native (https://github.com/pq-code-package/mldsa-native/pull/1119).

PR #459, opened 2026-09-17
RISCV: Add cosimulation test for RV32-IM backend

hanno-becker:
* Builds on #452 
* Builds on #455 
* Builds on #458 

This commit adds a QEMU-backed cosimulation infrastructure and CI run for the partial RV32-IM model introduced in #455.


riscv/proofs/cosim_runner.c
```diff
@@ -0,0 +1,452 @@
...
+        response_u32(state[i]);
+    }
+    response_send();
+}
+
+/* Return the trap metadata for an unexpected backend exception. */
+static void send_trap(void)
+{
+    response_reset();
+    response_string("TRAP mcause-");
```
aqjune-aws, line 375: 🟢 **Nit**

The emitted TRAP line (`TRAP mcause-<n> mtval-<n>`) does not match the protocol documented in this file's header comment at line 33 (`TRAP <mcause> <mtval> <status>`): the fields use `-` separators rather than being space-delimited values, and the `<status>` field is never emitted. This is only diagnostic output (the HOL side just concatenates `details`), so it's harmless, but the header comment and implementation should be reconciled — either emit `TRAP <mcause> <mtval> <status>` as documented, or update the doc comment to describe the `mcause-N mtval-N` form actually produced.

---

riscv/proofs/riscv.ml
```diff
@@ -254,10 +254,15 @@ let RISCV_CONV (decode_ths:thm option array) (ths:thm list) tm =
 let RISCV_BASIC_STEP_TAC =
   GEN_BASIC_STEP_TAC "RISCV" `riscv` `:riscvstate` RISCV_CONV ALL_TAC;;
 
-let RISCV_STEP_TAC (mc_length_th,decode_ths) subths sname
+let RISCV_STEP_TAC_WITH_REWRITES rewrite_ths
```
aqjune-aws: 🟡 **Suggestion**

`RISCV_STEP_TAC_WITH_REWRITES` is introduced with a `rewrite_ths` normalization parameter, but the only caller is `let RISCV_STEP_TAC = RISCV_STEP_TAC_WITH_REWRITES [];;` (line 277), which passes an empty list and thus reduces `normalize_tac` to `ALL_TAC` — behaviorally identical to the previous definition. Nothing in this PR (or elsewhere in the repo) exercises the non-empty-rewrites path. If this is groundwork for a later PR in the stack, that's fine, but as it stands the added parameterization is unused surface area; consider deferring it to the PR that actually needs it, or add a brief comment noting the intended future use.

---

aqjune-aws (commented):
Left a few AI-driven comments above. :)

.github/workflows/ci.yml
```diff
@@ -192,6 +192,48 @@ jobs:
           S2N_BIGNUM_SEMATEST_SEED=1 \
           "./$TARGET/proofs/simulator.native"
 
+  s2n-bignum-sematest-riscv:
```
aqjune-aws, line 195: Ideally if this is under `s2n-bignum-sematest` (line 85) it would be better, but I guess this was hard because this is using QEMU due to lack of RISC-V machine? (non-blocking comment)

riscv/proofs/cosim-runner
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+# Start the RV32IM cosimulation monitor under QEMU system mode.
+
+set -euo pipefail
+
+DIR="$(dirname -- "${BASH_SOURCE[0]}")"
+DIR="$(realpath -- "$DIR")"
+QEMU_COMMAND=${QEMU_RISCV32:-qemu-system-riscv32}
```
aqjune-aws, line 9: It would be great if the bash script fails explicitly when QEMU doesn't exist.

riscv/proofs/cosim.ld
```diff
@@ -0,0 +1,63 @@
+OUTPUT_ARCH(riscv)
```
aqjune-aws, line 3: Do you have any idea why a linker script was necessary?

riscv/proofs/cosim_runner.c
```diff
@@ -0,0 +1,452 @@
...
+    }
+}
+
+/* Compare two NUL-terminated strings for exact equality. */
+static int string_equal(const char *left, const char *right)
+{
+    while (*left != '\0' && *left == *right) {
+        left++;
+        right++;
+    }
```
aqjune-aws, line 146: It seems out of bounds buffer attack is possible. :) Maybe use `strcmp` instead of `string_equal`?

aqjune-aws (approved):
I have a few non-blocking comments; but it would be great if they could be addressed as a follow-up PR.

riscv/proofs/cosim_runner.c
hanno-becker: `left` and `right` are 0-terminated, so this is fine?

riscv/proofs/cosim-runner
hanno-becker: Thank you, the script now checks for QEMU :+1:

.github/workflows/ci.yml
hanno-becker: Yes. For RV64, that would work (e.g. we use native RV64 runners on mlx-native), but not for RV32.

riscv/proofs/cosim.ld
hanno-becker: This is needed because we're building a freestanding image, not an application binary running under syscall emulation (I don't think this is available for RV32).

riscv/proofs/cosim_runner.c
aqjune-aws: Sorry, you're right. I thought a case like `string_equal("ABC", "AB")` but since `"AB"` is also `\0`-terminated it stops.

